PR #10, opened 2026-09-11
fix(lspcmd): surface rename failures to the user

### LINKED ISSUE #2: LSP rename silently fails when the new name already exists in scope
LSP rename silently fails when the new name collides with an identifier that already exists in scope. The floating rename input just disappears and nothing happens: no error, no notification, no rename.

## Steps to reproduce

1. Open a Go file in a workspace with gopls running.
2. Put the cursor on a symbol and run the `rename` command.
3. Type a name that already exists in the same package or scope and press Enter.

**Actual:** the input box closes and nothing changes. No feedback.
**Expected:** a visible error explaining that the rename failed and why.

## Root cause

The rename flow calls `PrepareRename` once when the rename starts, which only validates that the cursor is on a renameable symbol. It does not validate the new name. When the actual rename request fails, the error is only logged and the floating input box is closed, so the user never sees it.

## Where to look

- `internal/ide/idelsp/lspcmd/rename.go` - the command handler and the floating handler that swallows the error on Enter.
- `internal/ide/idelsp/lsp.go` - `Rename` / `PrepareRename`, which proxy to gopls and return its raw error.
- `internal/ide/idelsp/go_test.go` - the table-driven e2e harness that runs against real gopls, including an existing rename happy-path case.

## Notes for contributors

This is a good first issue. The first step is investigation: add an e2e subtest that renames a symbol to a name that already exists in `testdata/` and record what gopls actually returns. Depending on the result, the fix is one of:

- gopls returns an error, so stop swallowing it and surface it to the user;
- gopls returns no error but an empty edit, so treat that as a failure with a user-visible message;
- the conflict is only detectable with a pre-flight check, so add it before applying the edit.

Lock the behavior in with a unit test at the `lspcmd` level so it does not depend on a live gopls.

MrBeldum:
## Summary

Fixes #2: LSP rename collisions (and other gopls rename failures) were only logged, so the floating rename input closed with no user-visible feedback.

This wires `browserapi.Notifications` into the rename floating handler and notifies `LevelError` when:

- gopls returns a rename error (e.g. name already declared in scope)
- the workspace edit is nil or contains no edits
- applying the workspace edit fails

Matches the existing error-notification pattern used by definition/hover/navigate handlers.

## Test plan

- [x] `go test ./internal/ide/idelsp/lspcmd/ -run 'TestRenameFloating|TestRenameHandler|TestApplyWorkspace|TestApplyEdits'`
- [x] Unit coverage for rename error, nil edit, empty edit, and apply-error notification paths
- [ ] Manual: rename a symbol to an existing name in-package and confirm an error notification appears


internal/ide/idelsp/lspcmd/rename.go
```diff
@@ -160,20 +164,53 @@ func (r *renameFloatingHandler) Handle(ev term.Event) (bool, bool) {
...
 	if err != nil {
 		r.log.Warn("rename apply", "err", err)
+		r.notifyRenameError(err)
 		return true, true
 	}
 	return true, true
 }
 
+func (r *renameFloatingHandler) notifyRenameError(err error) {
+	if r.notify == nil || err == nil {
```
ernestrc, line 185: We should strive to keep control flow explicit in the orchestrating function. Furthermore, `r.notify` should be guaranteed to not be nil at construction. What's the meaning of `r.notify` being nil? What's the production scenario that would pass a nil notification facility? It doesn't exist. If it's nil, we should let it panic: it's a programmer error. I would even move it at `renameFloatingHandler` so we enforce this invariant and let it crash the first time one of us constructs renameFloatingHandler with a nil notification facility.

If we agree that `r.notify` can't be nil, then the err != nil is redundant, at which point you can remove this helper altogether, and just inline, in notifyRenameError's call sites:

```go
if edit == nil || renameEditEmpty(edit) {
        _, _ = r.notify.Notify(browserapi.LevelError, "rename: no edits returned")
		return true, true
	}

...

if err != nil {
		r.log.Warn("rename apply", "err", err)
      	_, _ = r.notify.Notify(browserapi.LevelError, "rename: %s", err)
		return true, true
	}
```

internal/ide/idelsp/lspcmd/rename.go
```diff
@@ -160,20 +164,53 @@ func (r *renameFloatingHandler) Handle(ev term.Event) (bool, bool) {
...
+		return
+	}
+	_, _ = r.notify.Notify(browserapi.LevelError, "rename: %s", err)
+}
+
+// renameEditEmpty reports whether a workspace edit contains no text edits.
+// gopls can return a non-nil empty edit for some failure modes; treat that
+// the same as a failed rename so the user sees feedback.
+func renameEditEmpty(edit *semanticapi.WorkspaceEdit) bool {
+	if edit == nil {
```
ernestrc, line 195: This is currently redundant as it stands, because both call sites already checked that `edit` is indeed `nil.

internal/ide/idelsp/lspcmd/rename_test.go
```diff
@@ -73,7 +75,7 @@ func TestRenameFloatingHandlerEsc(t *testing.T) {
 		},
 	}
 
-	h := newTestRenameFloating(t, "oldName", lsp, &mockEditor{})
+	h := newTestRenameFloating(t, "oldName", lsp, &mockEditor{}, nil)
```
ernestrc, line 78: As discussed, don't pass nil here. Just pass a `nop` implementation of `browserapi.Notifications`. There's plenty in the codebase, if we were organized (which we ought to be!) we should have a `browsertest.NopNotifications()` implementation, and use in tests that we don't care about notifications.

Care to take a stab at this while we're at it?

```go
~/src/rune main*
❯ rg 'type nopNotifications struct\{\}'
internal/llm/llamaserver/e2e_test.go
258:type nopNotifications struct{}

internal/ide/idepkg/provisioning.go
99:type nopNotifications struct{}

internal/text/textrpc/client_server_test.go
1431:type nopNotifications struct{}

internal/ide/notifications.go
175:type nopNotifications struct{}
```

ernestrc:
Thanks for taking this one!

PR #56, opened 2026-09-12
Fix Markdown view tab reloads

### LINKED ISSUE #3: Markdown view tabs (:view file.md) cannot be reloaded
Opening a `.md` file with `:view` installs a read-only markdown tab. Reloading that tab fails: both `:reloadfile` and the filesystem watcher auto-reload report `failed to reload tab: cannot reload this content`.

## Steps to reproduce

1. Open a markdown file with `:view README.md`.
2. Change the file on disk.
3. Run `:reloadfile` (or wait for the FS watcher).

**Actual:** the notification `failed to reload tab: cannot reload this content` appears and the view keeps the stale content.
**Expected:** the rendered markdown updates to the new file contents.

## Root cause

Markdown view tabs are constructed with a `nil` tab closer, so they have no `workspace.FlusherCloser`. `Component.ReloadTab` rejects any tab whose closer is `nil` and returns `ErrInvalidReload` before any reload work happens. The markdown handler itself already supports content replacement via `SetComponent`, so the machinery for reloading exists, it just is not wired.

## Where to look

- `internal/text/component.go` - `newViewTab` (passes `nil` as the closer), `openFileTab`, `loadView` / `loadMarkdown`, and `ReloadTab`.
- `internal/handler/markdown/handler.go` - `SetComponent`, the content-swap mechanism a reload would use.
- `internal/text/flusher_closer.go` - `editorFlusherCloser`, the existing `FlusherCloser` implementation to mirror for the channel and event-loop dispatch patterns.

## Notes for contributors

The shape of the fix is a `FlusherCloser` for markdown view tabs: `Flush`/`ForceFlush` are no-ops, `Reload` re-reads and re-parses the file and swaps the component, and `LastFlush` reports the file modtime so the FS-watcher de-duplication keeps working. Once a non-nil closer is wired, `ReloadTab` needs no changes.

Constraints worth respecting: the read/parse happens off the event loop, but the UI mutation must be scheduled back onto the host event loop, and any goroutine must run under `debug.CapturePanicReport`. Tests should cover both `:reloadfile` and the FS-watcher path.

wthrajat:
Closes #3

- attach a flusher/closer to Markdown view tabs so they can participate in reloads
- read and parse the updated file off the event loop, then swap the rendered component on Rune's scheduler
- track the file modification time so filesystem watcher events behave consistently
- cover both `:reloadfile` and filesystem watcher reloads

#### testing

- `go test -race ./internal/text/... ./internal/ide/...`
- `make lint`

Verified `TestMarkdownViewReload`, `TestReload/reloads_markdown_view`, and the full affected suites pass.


internal/text/component.go
```diff
@@ -1601,33 +1603,53 @@ func (c *Component) newTab(
...
-		handler, err = c.loadMarkdown(file)
-	default:
+	if lang != "markdown" {
+		return nil, nil, false, nil
 	}
-	return
+
+	markdownComponent, modTime, err := c.loadMarkdown(file)
+	if err != nil {
+		return nil, nil, true, err
```
ernestrc: nit: if error != nil, the rest of params should be zero valued, i.e. `return nil, nil, false, err`

internal/text/markdown_flusher_closer.go
```diff
@@ -0,0 +1,160 @@
...
+	if m.closed {
+		m.mu.Unlock()
+		return nil, textapi.ErrInvalidReload
+	}
+	m.reloading = true
+	m.mu.Unlock()
+
+	result := make(chan error, 1)
+	go debug.CapturePanicReport(func() {
+		next, modTime, err := m.parent.loadMarkdown(m.uri)
```
ernestrc, line 88: Creating a new component+handler pair means that we lose search results and scroll position. From a UX perspective, imagine you're reading a markdown file, read-only mode, something changes it out of band (agent, git, etc.) and the content gets reset to the top and you lose your reading position. Not great.

I think we should [get the scroll position](https://github.com/unstablebuild/rune/blob/b82c2c21a18425f8e8dd1f3a9b1c5e5614d4a31f/internal/handler/markdown/handler.go#L318) (in the calling goroutine, not here, because it would cause a data race), and right after loading the markdown here, set the new scroll position. Same with the search results, although it might be wasteful to [re-search](https://github.com/unstablebuild/rune/blob/b82c2c21a18425f8e8dd1f3a9b1c5e5614d4a31f/internal/handler/markdown/handler.go#L359) just to re-create the state, then setting the scroll position, so we again guarantee there's no jerking of text up/down. Note that ordering matters: MaxSeekOffset and runSearch both depend on width/height, and the new component is built off-loop at size 0, it only gets sized inside SetComponent -> Resize. So restore has to happen after the swap: SetComponent(next) -> next.Search(query) -> next.SeekTo(offset).

Also two small additions to `markdown.Component` are needed since neither is exposed today:

• `SearchQuery()` string (the field is private; Search("") clears).
• `SeekTo(offset int)` clamping to [0, MaxSeekOffset()]. The clamp is what handles the file is now shorter case.

Please also add tests for the "file got smaller" offset is > MaxSeekOffset now, "file got bigger", search query that no longer matches → empty results, doesn't panic, etc as well.

internal/text/component.go
```diff
@@ -524,9 +524,11 @@ func (c *Component) openFileTabSync(
 }
 
 func (c *Component) newViewTab(
-	file workspaceapi.URI, h browserapi.Handler,
+	file workspaceapi.URI,
+	h browserapi.Handler,
+	closer workspace.FlusherCloser,
 ) *browser.Tab {
-	return c.newTab(file, c.iconFor(file), fileTabName(file), h, nil)
+	return c.newTab(file, c.iconFor(file), fileTabName(file), h, closer)
```
ernestrc, line 531: `closer == nil` used to be [the signal for `text.Component` to know when to return textapi.ErrInvalidSave](https://github.com/unstablebuild/rune/blob/b82c2c21a18425f8e8dd1f3a9b1c5e5614d4a31f/internal/text/component.go#L1199) so the user knows that it's trying to save a read-only markdown tab. With this change, all `write` calls to a read-only tab are now silently ignored. In #3 the "Flush should be a no-op" I meant, from an interface satisfaction's point of view, not from a user experience point of view, but I see how it could have been interpreted differently.

I think the best course of action would be to remove the `completedMarkdownOperation` helper and in `Flush` and `ForceFlush` simply return `nil, textapi.ErrInvalidSave` which is what `text.Component` returned for a markdown view before this changes.

internal/text/component.go
```diff
@@ -1601,33 +1603,53 @@ func (c *Component) newTab(
...
+	defer func() { _ = f.Close() }()
+
 	data, err := io.ReadAll(f)
 	if err != nil {
-		return nil, fmt.Errorf("read file: %w", err)
+		return nil, time.Time{}, fmt.Errorf("read file: %w", err)
 	}
-	_ = f.Close()
-	markdown, err := markdown.NewWithConfig(string(data), c.config.Markdown)
+	info, err := f.Stat()
```
ernestrc, line 1636: If a write lands between the two, `lastFlush` records the mtime of content that was never loaded. [`handleFSChange`](https://github.com/unstablebuild/rune/blob/b82c2c21a18425f8e8dd1f3a9b1c5e5614d4a31f/internal/ide/events.go#L90) then drops the incoming Write event as "user flushed file" and the preview stays stale permanently. 

Call `f.Stat()` before reading and it should avoid this.

ernestrc (commented):
Thanks so much for taking this!

ernestrc:
Also, please commit with `Signed-off-by: Author Name <authoremail@example.com>` to the DCO check passes.

internal/text/component.go
wthrajat: done!

internal/text/markdown_flusher_closer.go
wthrajat: okay that makes sense for a better UX. Pushed and also added the coverage!

internal/text/component.go
wthrajat: pushed (ordering is covered by a file double that exposes different mtimes before and after its first read)

internal/text/component.go
wthrajat: pushed!

wthrajat:
@ernestrc thanks for the review! Pushed nits, not resolving the comments myself, leaving that to you
(edit: resolved myseld after approval :p)

ernestrc (approved):

wthrajat:
@ernestrc can you re run the failed macos job? It probably got timed out since I cannot see an error and also that no change should trigger the failure.

PR #57, opened 2026-09-12
Fix visual-block delete clobbering the unnamed register

### LINKED ISSUE #1: vi: visual-block delete then paste only reinserts the last row
In vi mode, selecting a multi-line visual block with `<c-v>`, deleting it with `d` or `x`, and then pasting with `p` or `P` only reinserts the **last row** of the block. Vim pastes a block vertically: row N of the yanked block goes to row N below the cursor.

Yank-then-paste of a block already works. Only the delete-then-paste path is broken.

## Steps to reproduce

1. Open a file with several lines in vi mode.
2. Press `<c-v>`, extend the selection down a few rows and right a few columns.
3. Press `d`.
4. Move the cursor elsewhere and press `p`.

**Actual:** a single line is inserted, containing only the last row of the deleted block.
**Expected:** the whole block is reinserted vertically, one segment per line.

## Root cause

Two writers touch the unnamed register during a visual-block delete. The vi handler first copies the full block (rows joined with `\n`, metadata `BlockSelection`), which is correct. The delete then runs one edit per row, and each row's edit notifies the delete-clipboard subscriber, which overwrites the register with just that row's text. After the loop the register holds only the last row, so paste inserts a single line.

## Where to look

- `internal/text/delete_clipboard.go` - the `OnDidEdit` subscriber that overwrites the register per row.
- `internal/text/vi/handler.go` - `copySelectionForDelete` and the visual-mode `d`/`x` handler.
- `internal/cell/buffer.go` - `DeleteBlockContext`, the per-row edit loop.
- `internal/text/cursor.go` - `Cursor.Paste` / `InsertBlock`, which already handle blockwise pastes correctly when the register holds the joined text.

## Notes for contributors

Per the [bug-fix policy](https://github.com/unstablebuild/rune/blob/main/CONTRIBUTING.md#bug-fix-policy), start with a failing regression test. Table-driven cases worth covering: `d` then `p`, `x` then `p`, `d` then `P`, and a ragged block where rows have unequal length. The existing blockwise tests in `internal/text/vi/handler_test.go` are a good template. Make sure the black-hole register (`"_d`) behavior does not regress.

ikshantshukla123:
Visual-block `d`/`x` followed by `p`/`P` reinserted only the last row of
the block.

`copySelectionForDelete` writes the full block correctly, but the delete
runs one edit per row and each edit notifies the copy-on-delete
subscriber, which overwrites the register per row. The visual `d`/`x`
path now re-asserts the copied block once the delete completes, scoped
to block selections only. This also stops `"_d` on a block from touching
unnamed.

Fixes #1.

Regression test `TestVisualBlockDeleteThenPaste` (full-`Vi` level, where
the subscriber is installed): `d` then `P`, `x` then `p`, a ragged
block, and black-hole preservation : each checks register text, block
mode metadata, and the pasted buffer. The test fails on main with only
the last row in the register.


ikshantshukla123:
Note on scope: this resolves the two writer conflict in the visual d/x path (the explicit copy wins over the per row copyon delete writes) rather than restructuring the per row notifications in DeleteBlockContext/delClip. that keeps the blast radius to vi flows  the structural fix (block transactions for usage subscribers) would also touch Repeater and fits better as a follow up if you want it? @ernestrc

ernestrc:
Thanks for taking this! I gave you the green light on your approach but upon reviewing your PR, I saw an approach that avoids the need to correct the contents of the `unnamedRegister`. `delClip` [is installed with *Vi as a clipboard](https://github.com/unstablebuild/rune/blob/b0c51196edcab58e9f252ed3ecf0cd78946b2e81/internal/text/vi/vi.go#L103) implementation, allowing us to [intercept calls](https://github.com/unstablebuild/rune/blob/b0c51196edcab58e9f252ed3ecf0cd78946b2e81/internal/text/vi/vi.go#L291) and it's currently used to make sure undo/redo deletes are not copied to the clipboard, as well as backspace deletes in insert mode. 

We can add a new method to `viHandler` that then we can call in Vi.Copy:
```go
type viHandler interface {
        ...
>>	copySuppressed() bool
}
```

Then set a boolean in viHandler that changes the return value of this new method, so while we're doing block operations in [s/c visual selection, in handleVisualBlockChangeStart](https://github.com/unstablebuild/rune/blob/b0c51196edcab58e9f252ed3ecf0cd78946b2e81/internal/text/vi/handler.go#L1781) and [d/x visual selection](https://github.com/unstablebuild/rune/blob/b0c51196edcab58e9f252ed3ecf0cd78946b2e81/internal/text/vi/handler.go#L1930) we don't call the underlying clipboard's Copy, preventing the clobbering.

From a maintainability's point of view, I'd rather avoid an undesired behavior than amending the outcome of it.

ikshantshukla123:
> Thanks for taking this! I gave you the green light on your approach but upon reviewing your PR, I saw an approach that avoids the need to correct the contents of the `unnamedRegister`. `delClip` [is installed with *Vi as a clipboard](https://github.com/unstablebuild/rune/blob/b0c51196edcab58e9f252ed3ecf0cd78946b2e81/internal/text/vi/vi.go#L103) implementation, allowing us to [intercept calls](https://github.com/unstablebuild/rune/blob/b0c51196edcab58e9f252ed3ecf0cd78946b2e81/internal/text/vi/vi.go#L291) and it's currently used to make sure undo/redo deletes are not copied to the clipboard, as well as backspace deletes in insert mode.
> 
> We can add a new method to `viHandler` that then we can call in Vi.Copy:
> 
> ```go
> type viHandler interface {
>         ...
> >>	copySuppressed() bool
> }
> ```
> 
> Then set a boolean in viHandler that changes the return value of this new method, so while we're doing block operations in [s/c visual selection, in handleVisualBlockChangeStart](https://github.com/unstablebuild/rune/blob/b0c51196edcab58e9f252ed3ecf0cd78946b2e81/internal/text/vi/handler.go#L1781) and [d/x visual selection](https://github.com/unstablebuild/rune/blob/b0c51196edcab58e9f252ed3ecf0cd78946b2e81/internal/text/vi/handler.go#L1930) we don't call the underlying clipboard's Copy, preventing the clobbering.
> 
> From a maintainability's point of view, I'd rather avoid an undesired behavior than amending the outcome of it.

Thanks for more clarity on requirment @ernestrc
 Done switched to suppression as suggested  copySuppressed() on the viHandler interface, held during the block deletes in visual d/x (block-scoped) and handleVisualBlockChangeStart, gated in Vi.Copy next to the existing guards. The save/restore and its error path test are gone; the regression test is unchanged and green, plus a small test pinning that block c leaves unnamed alone.

ernestrc (approved):

ernestrc:
The reported workspace/explorer, startup race, OAuth EOF, and storage failover failures are addressed in follow-up #65 (with upstream OAuth fix unstablebuild/blue#3). These fixes are separate from the visual-block editing change here. Deterministic regressions and 50 repeated race-enabled runs of the affected tests pass locally; #65 CI is running.

PR #58, opened 2026-09-12
Consistently close the terminal

HigherOrderLogic:
Previously, running `exit` in a terminal only closed it when it's the focused handler. This patches always close the terminal when it exits.


internal/browser/component.go
```diff
@@ -1011,7 +1011,42 @@ func (c *Component) SetFocus(win Window) Window {
...
+		})
+		for _, id := range closing {
+			if bw, ok := c.findWindow(id); ok {
+				_ = bw.Close()
+			}
+		}
+	}
+	exit, handled = c.union.Handle(ev)
+	if exit {
+		c.RemoveWindowContent(c.focus())
```
ernestrc, line 1042: c.focus() is read after c.union.Handle(ev), so it is not necessarily the window that exited.

internal/ide/async_vte.go
```diff
@@ -399,6 +399,13 @@ func (av *asyncVTE) ClearPrimaryBuffer() bool {
 	return false
 }
 
+func (av *asyncVTE) Exited() bool {
+	if e, ok := av.real.(interface{ Exited() bool }); ok {
```
ernestrc, line 403: clever, but not necessary. An anonymous interface type assertion should be a last resort.

internal/browser/component.go
```diff
@@ -1011,7 +1011,42 @@ func (c *Component) SetFocus(win Window) Window {
 
 // Handle proxies events to either the underlying Tabs or WindowManager.
 func (c *Component) Handle(ev term.Event) (exit, handled bool) {
-	return c.union.Handle(ev)
+	if ev.Type == term.EventNone {
+		focusID := c.wm.Focus().ID()
+		for _, t := range slices.Clone(c.buffers) {
```
ernestrc, line 1016: No need to clone the slice here. It's causing allocations for no real benefit.

internal/browser/component.go
```diff
@@ -1011,7 +1011,42 @@ func (c *Component) SetFocus(win Window) Window {
...
+			content := w.Content()
+			if e, ok := content.(exitedReporter); !ok || !e.Exited() {
+				return
+			}
+			if exit, _ := content.Handle(ev); exit {
+				closing = append(closing, w.ID())
+			}
+		})
+		for _, id := range closing {
+			if bw, ok := c.findWindow(id); ok {
```
ernestrc, line 1035: We already had the window, why not store a slice of *browserWindow, so we don't need to do a map lookup here?

internal/browser/component.go
```diff
@@ -1011,7 +1011,42 @@ func (c *Component) SetFocus(win Window) Window {
 
 // Handle proxies events to either the underlying Tabs or WindowManager.
 func (c *Component) Handle(ev term.Event) (exit, handled bool) {
-	return c.union.Handle(ev)
+	if ev.Type == term.EventNone {
+		focusID := c.wm.Focus().ID()
+		for _, t := range slices.Clone(c.buffers) {
+			if t.free && t.Exited() {
+				t.Handle(ev)
```
ernestrc, line 1018: `browser.Tab.Handle` should be called when an event is dispatched to a particular `tui.Handler`.

This does what you intend it to do (*browser.Tab calling its parent browser and removing itself when it exits) but for a future reader of this code it might be quite hard to understand at first. I get that it's correct today, but we should strive to write code that it's easy to read for someone else in the future, no implicit contracts. In this same function, we're removing windows by calling `*browserWindow.Close`, removing tabs by calling `*browser.Tab.Handle` and removing tabs/content from a window again via `*browser.Component.RemoveWindowContent`.

ernestrc (commented):
Thanks for taking the initiative to fix this.

In principle, I'm not opposed to broadcasting term.EventNone from the workspace manager to all workspaces, and from each browser to all of its windows and free tabs. That does seem appealing, but I'm not convinced by the `exitReporter` interface. I think it overcomplicates the logic and isn't future-proof: a single intermediate wrapper that doesn't implement Exited breaks the chain of interface type assertions. The lack of tests makes this even more concerning.

There are also cases this doesn't address, such as extension-provided `tui.Handler` implementations installed via RPC, which use `handlerrpc.ClientStream[T]` and have the same [known limitation](https://github.com/unstablebuild/rune/blob/4f3615eaca6decfbda2a9e448f4a21c2b9bb29fd/internal/handler/handlerrpc/client_stream.go#L505). Maybe the approach of sending `term.EventNone` to force a call to `.Handle`, thereby prompting collections to remove exiting children, has run its course.

The vte already knows when the shell exits (where we set `e.exit` in `vte.Handler`), and it already talks outward to the browser: `browser.TabManager` is passed into `vte.NewHandler`. We can declare a new interface in the `vte` package with just the method vte needs, and `browser.Component` satisfies it directly which it's already what gets passed as `TabManager` today. 

```go
// declared in package vte
type TabExiter interface {
	OnTabExit(uri workspaceapi.URI) bool
}
```

That gives us a compile-checked contract instead of an assertion chain, identity comes from the URI at call time, and it works the same for free tabs and for windows in non-focused workspaces; `workspaceTabManager.SetTabName` already hops `scheduleNextTick` and updates a workspace that isn't in focus.

Also, please always add tests that cover, at a minimum, the bug we're trying to fix. Otherwise, we risk reintroducing the issue in the future.

HigherOrderLogic:
Addressed in 2e0ec4da:

- Dropped the `exitedReporter` assertion chain and the EventNone broadcast entirely (including the workspace fan-out).
- `vte` now declares `TabExiter` and the handler asks its own `TabManager` to drop it the moment the pty child exits, hopping `ScheduleNextTick` so `OnTabExit` runs on the host event loop.
- `browser.Component` implements `OnTabExit(uri)`: tabs are removed wherever they live (free or attached), and direct window content (`terminalnew`, `terminalneworsplit`) is swapped to the next free tab or start content, mirroring tab removal. `tabNameAliaser.OnTabExit` resolves pty URIs to the key the terminal was created under, same as `SetTabName`.
- Kept the last-tile exit restore in `Component.Handle`: the IDE still discards the window manager's exit signal, and this covers non-vte handlers exiting as the last tile.
- Tests: e2e regression for the reported bug (`terminalneworsplit` + `exit`), `TestHandlerCallsTabExiterOnPtyExit` (real pty), `TestComponentOnTabExit` (free tab, attached tab, direct window content), and alias-resolution cases in `TestTabNameAliaser`.

The extension RPC handler limitation is unchanged as discussed.

internal/browser/component.go
HigherOrderLogic: `exit == true` from `union.Handle` means that the window has exited successfully, and there are no other windows, meaning we're at the homepage of the editor.

internal/term/vte/component.go
```diff
@@ -174,10 +176,19 @@ func (t *Component) triggerBell() {
...
-// Run must be called in a separate goroutine to start processing incoming
-// data from the pty master.
+type TabExiter interface {
+	OnTabExit(uri workspaceapi.URI) bool
+}
+
 func (t *Component) Run(publisher browser.EventPublisher) error {
-	return t.run(publisher)
+	err := t.run(publisher)
+	if ex, ok := t.tm.(TabExiter); ok {
```
ernestrc, line 185: I've verified it manually and now `sleep 2 && exit` on a vte and switching to another tile doesn't close the window anymore. I suspect there's an intermediate interface here that you forgot to update it to satisfy `TabExiter`; my biggest concern in your prior design/approach.

Add `OnTabExit` to `browser.TabManager`. This way the compiler will tell us if one of our TabManager implementations is missing OnTabExit, now and in the future.

internal/term/vte/component.go
```diff
@@ -174,10 +176,19 @@ func (t *Component) triggerBell() {
...
+		uri := t.uri
+		if !t.cfg.ScheduleNextTick(func() { ex.OnTabExit(uri) }) {
+			ex.OnTabExit(uri)
```
ernestrc, line 188: if `ScheduleNextTick` ever returns false, this would cause a race. If we can't schedule events, we have bigger fish to fry than exiting a vte. Dropping the call is acceptable here.

internal/browser/component.go
```diff
@@ -1011,7 +1011,46 @@ func (c *Component) SetFocus(win Window) Window {
...
-	return c.union.Handle(ev)
+	exit, handled = c.union.Handle(ev)
+	if exit {
+		c.RemoveWindowContent(c.focus())
+		exit = false
+	}
+	return
+}
+
+func (c *Component) OnTabExit(uri workspaceapi.URI) bool {
```
ernestrc, line 1022: The linter is failing here. OnTabExit needs a comment.

internal/term/vte/component.go
```diff
@@ -174,10 +176,19 @@ func (t *Component) triggerBell() {
 	}
 }
 
-// Run must be called in a separate goroutine to start processing incoming
-// data from the pty master.
```
ernestrc, line 178: We shouldn't remove this comment: it's causing the linter to fail

internal/term/vte/handler_test.go
```diff
@@ -294,6 +294,60 @@ func TestHandlerPublishesEventOnPtyExit(t *testing.T) {
...
+	return true
+}
+
+func (r *tabExiterRecorder) ExitedURIs() []workspaceapi.URI {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	return append([]workspaceapi.URI(nil), r.exited...)
+}
+
+func TestHandlerCallsTabExiterOnPtyExit(t *testing.T) {
```
ernestrc, line 316: `TestHandlerCallsTabExiterOnPtyExit` only runs with `make test-e2e` because it's not a hermetic test (`//go:build e2e`). This might not be intuitive for you, so that's my bad. Normally when we add a test for such a small change, we should have it fail first, then update/fix, then see that i passes (TDD).

This means that `make test` covers none of your changes. There's no hermetic test of `browser.Component.OnTabExit`, and nothing tests the wiring, which is the part that's broken. The more loosely coupled the logic in `browser.Component` is (i.e. anonymous interface assertions), the more and better test coverage we must have.

internal/browser/component.go
ernestrc: I think making these kind of assumptions here is what could get us into subtle bugs territory. The semantics of `exit` are somewhat loose, specially for collections like union and wm. 

In any case, `OnTabExit` already gets called unconditionally, so I think this branch is redundant. If it isn't add the test that proves that this branch is necessary, so anyone mistakenly deeming it as unnecessary in the future doesn't delete it by mistake and re-introduces the bug.

internal/browser/component.go
```diff
@@ -1011,7 +1011,46 @@ func (c *Component) SetFocus(win Window) Window {
...
+	for _, t := range c.buffers {
+		if t.uri.String() == uri.String() {
+			return c.RemoveTab(t)
+		}
+	}
+	var exited *browserWindow
+	c.wm.Iterate(func(w thandler.Window) {
```
ernestrc, line 1029: no need to iterate on c.wm windows, which gets you only `tui.Handler`. You can iterate directly on the browser windows, bypassing the need for the first interface assertion:

```go
for _, bw := range c.windows {
		h, _ := bw.Content()
		u, ok := h.(uriHandler)
		if !ok || u.URI().String() != uri.String() {
			continue
		}
        if u.URI().String() == uri.String() {
           	exited, _ = c.findWindow(w.ID())
        }
		break
	}
```

The only interface assertion needed now is `URIHandler`, which I'd prefer it if it's defined as a private ifc elsewhere in this file, rather than type asserting an anonymous ifc.

internal/browser/component.go
ernestrc: In Go, docstrings must start in the form of:
```go
// Method bla bla
func (c comp) Method
```

So here you need 
```go
// OnTabExit drops the terminal with the given uri.
func (c *Component) OnTabExit(uri workspaceapi.URI) bool {
```

ernestrc (approved):

PR #64, opened 2026-09-13
Fix file explorer expand/collapse for symlinks and gitignored folders

### LINKED ISSUE #53: File explorer: symlinked directories and childless folders cannot be expanded or collapsed
Reported by a user:

> Also, idk if it's the .gitignore but i can't expand/collapse .gitignored folders in the file explorer

Investigation turned up three distinct defects in the file explorer that all present as "this folder does not expand or collapse".

## 1. Symlinked directories render as files

`FileSystem.ReadDir` is lstat-based, so a symlink pointing at a directory arrives with `IsDir() == false`. `component/fileexplorer/tree.go:readChildren` used that value verbatim, with two consequences that together explain the report:

- The row is built with `isDir: false`, so it renders without a trailing `/`, gets the file glyph, and `<Enter>` routes to "open in editor" instead of toggling. The folder can be neither expanded nor collapsed.
- The same unresolved value is passed to the ignore matcher, so a dir-only gitignore pattern (`node_modules/`, `build/`, `target/`) does not match it and the entry stays visible. The user sees a folder they listed in `.gitignore`.

Note the matcher behavior is correct here — git treats a symlink as a blob, not a tree:

```
$ printf 'linked/\n' > .gitignore && ln -s real linked
$ git check-ignore -v linked ; echo $?
1
$ git status --porcelain
?? .gitignore
?? linked
```

Reproduced against a real file scheme; the tree rendered `linked` as a leaf next to `real/`.

**Fix:** resolve symlink entries with `Stat` for directory-ness, while still matching ignore patterns against the unresolved entry so git semantics are preserved. Broken links stay files.

## 2. Directories with no visible children latch open

`parseViewTree` inferred expansion state purely from the buffer: a directory was "expanded" iff at least one child row was parsed under it. A directory that renders no rows at all — empty on disk, or with every entry hidden by `Config.Ignore` — therefore always parsed back as collapsed. `ExpandNodeAt` re-expanded it on every `<Enter>`, so the open-directory glyph latched on and the row could never be collapsed.

This is easy to hit in this repository:

- `internal/handler/fileexplorer/` contains only `.handler.go.swp` and `.mode.go.swp` (both ignored)
- `ide/idelsp/symbolresolve/testdata_py_pkg/src/` contains only `mypkg.egg-info`

It also compounds defect 1: a symlinked directory that gitignore hides the contents of hits both at once, which is probably why the report conflated them.

**Fix:** when the view has no child rows and the base node has no children either, take the expanded flag from the base tree. The narrow `len(base.children) == 0` condition preserves the "user deleted every child row = intentional delete" semantics for real directories.

## 3. Expand failures are silent (open)

`Component.ExpandNodeAt` swallows the `readChildren` error and returns a no-op, so a directory the user cannot read gives no feedback at all — `<Enter>` simply does nothing. Verified with a `chmod 000` directory. Not gitignore-related, and left unfixed for now; worth surfacing through the host error channel.

## Scope

1 and 2 are fixed with regression tests in `internal/component/fileexplorer` (`TestSymlinkedDirectoryIsBrowsable`, `TestToggleDirectoryWithoutVisibleChildren`). 3 is still open.

sanjay7178:
## Summary
- Treat symlink-to-directory entries as folders by resolving directory-ness with `FileSystem.Stat`, while still matching ignore patterns against the unresolved `ReadDir` entry so git blob semantics stay intact. Broken links remain files.
- Let directories with no visible children collapse. `parseViewTree` used to infer expansion only from child rows, so empty folders (and folders whose contents were all ignored) re-parsed as collapsed and `<Enter>` could never close them.
- When every child of a folder is ignored (e.g. `test1/*` in `.gitignore`), still list those children on expand so the folder is browsable. Mixed directories continue to hide ignored noise such as `.git/` and `*.swp`.

Fixes https://github.com/unstablebuild/rune/issues/53

Silent expand failures on unreadable directories (`chmod 000`) are unchanged; `ExpandNodeAt` still swallows `readChildren` errors.


internal/component/fileexplorer/tree.go
```diff
@@ -75,26 +76,52 @@ func readChildren(
 	if err != nil {
 		return err
 	}
-	children := make([]*node, 0, len(entries))
+	visible := make([]*node, 0, len(entries))
+	hidden := make([]*node, 0, len(entries))
```
ernestrc, line 80: Using `cap` to pre-allocate is only useful when we know the lower-bound of items that will get appended to a slice. 

Most directories won't have anything hidden so this is performing an extra alloc for no real benefit.

internal/component/fileexplorer/tree.go
```diff
@@ -75,26 +76,52 @@ func readChildren(
...
+			continue
+		}
+		visible = append(visible, child)
 	}
-	sortChildren(children)
-	n.children = children
+	// Keep those children so the folder stays browsable; mixed
+	// directories still hide ignored noise.
+	if len(visible) == 0 {
+		visible = hidden
```
ernestrc, line 102: This is very counter-intuitive from a UX perspective: it silently disables the gitignore filter for any directory whose entries are all ignored. Then if I create a file that's not gitignored, the previous files disappear and only the new one appears. 

If the goal is "a non-ignored folder must always be browsable," this should be an explicit, consistent rule (i.e render ignored entries always (dimmed), or track a per-node "show ignored" flag toggled by the user on expand), rather than a count-based fallback baked into `readChildren`, which also feeds `baseTree` and therefore `computeChangeSet`.

internal/component/fileexplorer/fileexplorer.go
```diff
@@ -785,17 +785,25 @@ func (c *Component) parseViewTree() *node {
...
-					c.expanded = true
-				}
-				markExpanded(c)
+		for _, child := range n.children {
+			if !child.isDir {
+				continue
+			}
+			if len(child.children) > 0 {
+				child.expanded = true
+			} else if base := c.findNodeByID(c.baseTree, child.id); base != nil &&
```
ernestrc, line 802: before calling `findNodeByID` we must check that `child.id` is not 0: for a brand-new, unflushed directory row (id == 0) this resolves to the base tree root, which is always `expanded: true`.

internal/component/fileexplorer/tree.go
```diff
@@ -75,26 +76,52 @@ func readChildren(
...
+	}
+	if e.Type()&os.ModeSymlink == 0 {
+		return false
+	}
+	info, err := fs.Stat(uri.Path())
+	if err != nil {
+		return false
+	}
+	return info.IsDir()
+}
```
ernestrc, line 123: This symlink following is correct for browsing, but we haven't solved it for the edit/flush pipeline. Try adding a test that expands a symlink directory, containing a `inside.go` file, then rename the row to something else, and run `DryFlush`:

```
op: rename linked -> linked2
op: move /project/linked/inside.go -> /project/linked2/inside.go
```

Now `applyOperation` (fileexplorer.go:662-669) uses `c.fs.Stat`, which follows the link, sees `IsDir()`, and does `MkdirAll("/project/linked2")`. Net effect after flush: a real directory linked2 + inside.go physically moved out of the symlink target, and the original linked symlink still on disk.

Same concern for OpDelete (:658): Remove on the symlink unlinks it, but child delete ops would first delete real files in the target.

At minimum this needs a test covering rename/delete of a symlinked directory, and probably a guard (skip dir-style handling when Lstat says symlink, or refuse to expand-and-edit symlinks).

internal/component/fileexplorer/tree.go
```diff
@@ -75,26 +76,52 @@ func readChildren(
...
 			name:   e.Name(),
 			uri:    childURI,
-			isDir:  e.IsDir(),
+			isDir:  dirEntryIsDir(fs, e, childURI),
 			depth:  n.depth + 1,
 			parent: n,
 		}
-		children = append(children, child)
+		// Match the unresolved DirEntry so a dir-only gitignore
+		// pattern does not hide a symlink (git treats it as a blob).
```
ernestrc, line 92: I think it's worth documenting this special case in `config.Ignore`.

ernestrc (commented):
Sorry for taking so long to review, the fileexplorer is complex and it needs special attention.

PR #65, opened 2026-09-13
Fix CI races in workspace saves, login, and storage failover

ernestrc:
## Summary
Fixes four underlying CI failures reported on #57 (without changing that fork PR):

- Defer filesystem conflict checks through save/reload UI completion, then recheck disk rather than dropping external changes. Release deferred callbacks if scheduling is rejected.
- Preserve external writes between save rename and reopen: only adopt a changed timestamp when the reopened content matches the actual saved bytes. Cover concurrent buffer edits and stale-save protection.
- Use the shared queued test scheduler so workspace installation cannot race IDE initialization.
- Reconnect idle gRPC storage followers and close abandoned connections; prove failover still works when the leader goodbye is lost.
- Pin the OAuth callback flush fix from https://github.com/unstablebuild/blue/pull/3 and force the callback/shutdown ordering in the Rune regression.

## Validation
- Deterministic regressions failed before each fix.
- Full make test and make lint pass; zero lint issues.
- Combined affected-test stress suite passes 50 repetitions under -race.
- Added rename/reopen regressions pass another 30 repetitions under -race.
- Serial CI-style make test also passed before the final rename/reopen follow-up; an uncached repeat is running on the complete patch.
- make generate succeeds; unrelated generated copyright-header drift excluded.
- Hosted Linux/macOS checks are running on the latest commit.

## Dependency
Blue PR https://github.com/unstablebuild/blue/pull/3 is the companion upstream fix and its complete CI matrix passes. Rune pins its immutable commit as a Go pseudo-version; no local replace or release publication is required.


ernestrc:
Local validation update: the combined affected-test stress run passed 50 repetitions per test under -race (IDE 66.5s, firstmover 30.8s, apiclient 3.0s). Full make test passed in the commit hook; make lint reports zero issues. The companion Blue PR #3 passed its complete Linux/macOS stable/oldstable CI matrix. Rune's hosted checks and an additional serial make test run are in progress.

ernestrc:
Pushed 74d91e69 after review exposed a related rename/reopen window: an external edit could be adopted as Rune's saved timestamp, causing the deferred watcher to skip it. Added workspace and handler regressions (including pending user edits), and verify changed timestamps against the saved swap digest before accepting them. Normal unchanged-mtime saves do not incur another read.

Validation: new regressions pass 30 race-enabled repetitions; lint reports zero issues; the full make test commit hook passed again. The earlier serial make test GOTESTFLAGS='-race -timeout 300s -p 1' also passed. Hosted CI is rerunning on the complete patch.

PR #70, opened 2026-09-15
Fix Rune failing to start without graphical libraries installed

### LINKED ISSUE #66: Allows headless environment (i.e TUI, no OpenGL, no X.org, etc) to use Rune
### Problem

Currently the UI looks like it should supposed to be runnable as TUI under headless environment but apparently it requires graphical environment as I just found out which is a bummer.

### Proposed behavior

Start successfully as TUI in headless environment.

### Area

GUI and rendering

### Alternatives considered

_No response_

### Additional context

_No response_

### Checks

- [ ] I would be willing to work on this.

ernestrc:
A Rune binary listed libX11.so.6 in DT_NEEDED, so the dynamic loader refused to start it on a server with no display stack — before main() ran, which made the mode flags irrelevant. The ebiten fork now resolves Xlib through dlopen when it opens a window, leaving libc and libm as the only shared libraries the binary needs.

With that out of the way, promote the hidden --hardcore flag to --tui and also add --headless, which serves a machine's workspaces on the Rune network with no UI so everyone has access to this.

Fixes #66.


cmd/rune/headless.go
```diff
@@ -0,0 +1,250 @@
...
+	case <-ctx.Done():
+		return ctx.Err()
+	case err := <-session.Done:
+		if err != nil {
+			return err
+		}
+	}
+
+	user, ok, err = client.AccountStatus(ctx)
+	if err != nil || !ok {
```
wthrajat: This branch hides a broken cached token. `AccountStatus` returns an error only when the token cannot be decoded, but here that error prints `Login successful` and returns nil so the node keeps joining with bad auth. Return the error first and keep the fallback only for `!ok`.

cmd/rune/headless.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	fmt.Fprintf(&b, " (%s)\n", headlessPlanLabel(u.Role))
+	return b.String()
+}
+
+func headlessPlanLabel(role auth.Role) string {
+	switch {
+	case role == auth.RoleOneOff:
+		return "One-off"
+	case role >= auth.RolePaid:
```
wthrajat, line 227: this label shows Admin and SuperAdmin as `Rune Pro`. `auth/user.go` keeps `RoleOneOff` last on purpose and says to match roles exactly instead of using `>=`
intentional?

cmd/rune/main.go
```diff
@@ -509,8 +508,13 @@ func run() int {
 		return code
 	}
 
-	var mu sync.Mutex
 	ctx := context.Background()
+
+	if *flagHeadless {
```
wthrajat, line 573: `--headless` silently wins when combined with `--tui` or `--gui`
if both flags are set in a script it gets a headless node with no warning while file args are also ignored
maybe we can error out when more than one mode flag is set

.github/workflows/test-linux.yml
```diff
@@ -36,6 +36,56 @@ jobs:
...
 
+      # Graphical libraries may be linked against at build time, but the
+      # dynamic loader resolves DT_NEEDED before main(), so any of them
+      # listed here makes rune unstartable on a machine without a display
+      # stack — whatever mode flags the user passed.
+      - name: Assert no graphical library is needed at load time
+        run: |
+          objdump -p ./bin/rune | awk '$1 == "NEEDED" { print $2 }' > needed.txt
+          cat needed.txt
+          if grep -E '^lib(X|GL|EGL|GLES|wayland|xkbcommon)' needed.txt; then
```
wthrajat: nit: this check misses `libOpenGL`

.github/workflows/test-linux.yml
```diff
@@ -36,6 +36,56 @@ jobs:
...
+
+            # The alternate-screen switch is the first thing the TUI
+            # writes, so seeing it means the editor came up on a machine
+            # with no display stack at all.
+            out=$(TERM=xterm-256color timeout -s TERM 20 \
+              script -qec "/rune/rune --tui -d /tmp/runedata" /dev/null \
+              </dev/null 2>&1 || true)
+            want "rune --tui" "[?1049h" "$out"
+
+            out=$(timeout -s TERM 20 /rune/rune --headless -d /tmp/runedata 2>&1 || true)
```
wthrajat: both the test runs share `/tmp/runedata`. The TUI run is killed by timeout and can leave locks behind, so the headless run starts on dirty state and can flake
we can maybe a fresh data dir for each run

cmd/rune/headless.go
```diff
@@ -0,0 +1,250 @@
...
+	return b.String()
+}
+
+// waitForShutdownSignal blocks until the service manager (or the
+// operator) asks the node to stop.
+func waitForShutdownSignal() {
+	ch := make(chan os.Signal, 1)
+	signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
+	defer signal.Stop(ch)
+	sig := <-ch
```
wthrajat: afaik this wait never watches the context since the `runHeadless` takes `ctx` for login and join, but shutdown blocks only on signals so cancel never stops it
maybe we can pass `ctx` in and select on both

wthrajat:
@ernestrc I left some nits/questions

cmd/rune/headless.go
ernestrc: I wasn't expecting your review, thanks for taking the time! I caught this and fixed it in my last push.

.github/workflows/test-linux.yml
ernestrc: good catch!

.github/workflows/test-linux.yml
ernestrc: Fair point! Tests should be as hermetic as possible.

cmd/rune/headless.go
ernestrc: RoleOneOff doesn't make sense now that Rune is open source, so it'll be removed soon.

ernestrc: To answer your question, intentional.

cmd/rune/main.go
ernestrc: Good call!

cmd/rune/headless.go
ernestrc: the ctx is `context.Background` in main.go, but still, you're right and we should not be making these kind of assumptions from this function.

wthrajat:
@ernestrc feel free to resolve all threads.

PR #79, opened 2026-09-16
Fix auto-reload for files opened outside the workspace root

AniketR10:
https://github.com/user-attachments/assets/5240a12a-f728-4ba3-bbcf-9ccaa2428a57

/tmp/outside.txt, a file outside the workspace root, reloads automatically when it's changed externally.


AniketR10:
@ernestrc please reivew

internal/ide/out_of_root_watch.go
```diff
@@ -0,0 +1,240 @@
...
+		ex:     e,
+		scheme: scheme,
+		root:   e.workspaceURI,
+		uiMu:   uiMu,
+		dirs:   make(map[string]*outOfRootDir),
+	}
+	if err := e.comp.SubscribeEvents(outOfRootTabEvents, w); err != nil {
+		return err
+	}
+	e.outOfRootTabs = w
```
ernestrc: We should try to keep ex initialization to its constructor/initializer. Keeping all initialization to one function/method helps readability and allows us to enforce invariants that otherwise we can't enforce.

```go
) (e *ex, err error) {
	e = new(ex)
    /* ... */
	e.outOfRootTabs = &outOfRootWatcher{
		ex:     e,
		scheme: m,
		root:   e.workspaceURI,
		uiMu:   locker,
		dirs:   make(map[string]*outOfRootDir),
	}
	err = e.comp.SubscribeEvents(outOfRootTabEvents, e.outOfRootTabs)
	if err != nil {
		return
	}
	e.outOfRootTabs = w
	return
}
```

It looks like we'd need to add a new parameter to `newEx`, `sync.Locker`. I know the constructor has way too many parameters already, but the solution to that is breaking down `ex` into smaller structs, and distribute responsibilities accordingly, rather than adding methods that conditionally initialize certain parts of it.

Even better:

```go
) (e *ex, err error) {
	e = new(ex)
    /* ... */
	e.outOfRootTabs = newOutOfRootWatcher(...)
	err = e.comp.SubscribeEvents(outOfRootTabEvents, e.outOfRootTabs)
	if err != nil {
		return
	}
	e.outOfRootTabs = w
	return
}
```

Then in `out_of_root_watch.go`:
```go
func newOutOfRootWatcher(...) *outOfRootWatcher {
  return &outOfRootWatcher{
		ex:     e,
		scheme: m,
		root:   e.workspaceURI,
		uiMu:   locker,
		dirs:   make(map[string]*outOfRootDir),
	}
}
```

internal/ide/ex.go
```diff
@@ -2931,6 +2932,9 @@ func (e *ex) Close() (ret error) {
 	if err := e.comp.Close(); err != nil {
 		ret = multierror.Append(ret, err)
 	}
+	if e.outOfRootTabs != nil {
```
ernestrc: For example here `e.outOfRootTabs` should never be nil in prod. (context: https://github.com/unstablebuild/rune/pull/79/changes#r4030158099)

internal/ide/out_of_root_watch.go
```diff
@@ -0,0 +1,240 @@
...
+// watch arms the directory watch off the UI goroutine, since on remote
+// workspaces every Watch is an RPC round trip. It owns StopWatch, so a watch
+// canceled while still arming is released once Watch returns.
+func (w *outOfRootWatcher) watch(ctx context.Context, dir string) {
+	ch := make(chan schemeapi.EventInfo, 64)
+	id, err := w.scheme.Watch(dir, ch,
+		schemeapi.Create, schemeapi.Write,
+		schemeapi.Remove, schemeapi.Rename)
+	if err != nil {
+		w.ex.log(log.WarnLevel, "watch out-of-root tabs in %s: %v", dir, err)
```
ernestrc: We should return this error to the caller, otherwise it's unaware that there's no watch for a particular dir, so a dir remains in `w.dirs` and `add` only fires the goroutine when the dir key is absent, so after one failure every later tab in that directory silently gets no reload, forever.

internal/ide/out_of_root_watch.go
```diff
@@ -0,0 +1,240 @@
...
+		if w.isOutOfRoot(ev.URI) {
+			w.add(ev.URI)
+		}
+	case textapi.EventTypeClose:
+		w.remove(ev.URI)
+	}
+	return false
+}
+
+func (w *outOfRootWatcher) isOutOfRoot(uri workspaceapi.URI) bool {
```
ernestrc, line 137: If cwd is `/tmp/ws` (really `/private/tmp/ws`) plus a tab opened at the realpath, yields `rel == "../../private/tmp/ws/f.go"`, it is classified out-of-root and watched.

But the workspace recursive watcher also reports that same path and `comp.Resource` matches the tab. One save then produces two reloads, or two stacked "changed on disk" prompts. Consider `filepath.EvalSymlinks` on both sides before the containment test. Add a symlinks test to your test suites.

internal/ide/out_of_root_watch.go
```diff
@@ -0,0 +1,240 @@
...
+			return
+		case fsev, ok := <-ch:
+			if !ok || ctx.Err() != nil {
+				return
+			}
+			tab, ok := w.tab(dir, filepath.Base(fsev.URI().Path()))
+			if !ok {
+				continue
+			}
+			dispatchFilesystemEvent(w.ex, w.uiMu, vctrl.NopMatcher(false),
```
ernestrc: `dispatchFilesystemEvent` also calls `ex.comp.DispatchEvent`, so out-of-root FS changes now reach every subscriber for the first time, i.e. `idelsp.Manager` broadcasts `workspace/didChangeWatchedFiles`
to all running LSP servers, plus rune-agent, history, and really all extensions. I'm particularly concerned with LSP servers, because we don't control that code and `idelsp.Manager` is poorly positioned to make a call. 

The blast radius is too big so I think we should consider extracting the event mapping logic from `dispatchFilesystemEvent` into a helper and calling `handleFSChange` directly here.

ernestrc (commented):
I'd push back on using `w.scheme.Watch` for files outside of the workspace. The reason is subtle, but the stakes are high:[ on macOS the FSEvents watcher is (directory) recursive and just filters out events post-hoc](https://github.com/unstablebuild/notify/blob/a40652382ffdda6a8d5ccae0053b17a2f52aa7c1/watcher_fsevents.go#L72). If a file changes in $HOME, we'd be watching and streaming all the changes to files under $HOME until that file is closed. Our logic here wouldn't see any of these events, but it would hurt runtime performance quite a bit.

We can't use something like `runtime.GOOS` here either, because the watcher implementation lies with the workspace, and the workspace could be a remote workspace running on linux while the local is macOS.

I think the fix is to stop subscribing and instead poll for file changes on a tightly controlled loop (by calling `.Stat` on the `workspaceapi.File`), designed to reduce overhead and limit latency to something reasonable. Tuning this loop will require manual testing and benchmarking.

What it costs: one stat per out-of-root file per tick, so 20 tabs at 2s is 10 stats/sec, versus an FSEvents stream potentially over $HOME on macOS, but otherwise fine on linux. On remote workspaces it's a bit more costly, specially on linux which would not be a problem with your current approach:replaces N server-streaming gRPC watches (workspacerpc.Server.Watch) with N unary Stats; I think there's an interface type assertion we can do here to figure out if the workspace is remote, and use a longer interval or focus-only polling when the scheme is remote. Or we could disable this feature, altogether if the workspace is remote.

AniketR10:
@ernestrc have fixed the issue please reivew

internal/ide/out_of_root_watch.go
```diff
@@ -0,0 +1,284 @@
...
+	case prev.exists && !cur.exists:
+		return schemeapi.Remove, true
+	case cur.exists && (!prev.modTime.Equal(cur.modTime) || prev.size != cur.size):
+		return schemeapi.Write, true
+	}
+	return 0, false
+}
+
+func (w *outOfRootWatcher) stat(uri workspaceapi.URI) fileState {
+	state, _ := w.statErr(uri)
```
ernestrc, line 255: A transient failure records now records`exists: false` (fileState's zero value for `exists bool` is false) as the baseline. The next successful poll sees `false -> true`, `fileChange` returns `schemeapi.Create`, and `handleFSChange` acts on it. On a clean buffer that is a spurious reload plus a "changed on disk" notification. On a dirty buffer it is a modal prompt telling the user the file was created on disk, when nothing touched it.

You might think that a `Stat` returning an error is pretty unlikely, specially since we only run this logic on local workspaces, but we shouldn't write software that works in *almost* all cases. Generally speaking we should never hide errors like this.

internal/ide/out_of_root_watch.go
```diff
@@ -0,0 +1,284 @@
...
+		if w.enabled && w.isOutOfRoot(ev.URI) {
+			w.add(ev.URI)
+		}
+	case textapi.EventTypeClose:
+		w.remove(ev.URI)
+	}
+	return false
+}
+
+func (w *outOfRootWatcher) isOutOfRoot(uri workspaceapi.URI) bool {
```
ernestrc, line 137: There's a little side-effect of this check asking the question "is this file outside of the workspace": for the home workspace, which we don't have a recursive watcher installed, we're going to watch files outside of the home directory, but we have nothing watching files inside.

now that we have this new poll mechanism, we should use it for the home workspace and change this `isOutOfRoot` with a new `hasRecursiveWatcher`, which takes this special `h.empty`/home workspace consideration. This is out of scope for this PR, but I thought I'd give you the opportunity for a follow up.

internal/ide/out_of_root_watch.go
```diff
@@ -0,0 +1,284 @@
...
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			w.pollOnce(ctx)
+		}
+	}
+}
+
+// pollOnce stats every tab without holding any lock, then reports the
+// changes under uiMu, as the workspace watcher does.
```
ernestrc, line 201: the comment is misleading, `pollOnce` calls `report` for every tab, and `report` takes `uiMu` before it
decides whether anything changed. In the steady state where nothing has happened, 20 out-of-root tabs means 20 acquisitions of the process-wide event loop/UI lock every 2 seconds for no reason.

Compute `fileChange` under `w.mu`, return early when `!changed`, and take `uiMu`
only for real transitions:

```go
w.mu.Lock()
tab, ok := w.tabs[uri.String()]
if !ok {
	w.mu.Unlock()
	return
}
flag, changed := fileChange(tab.state, state)
tab.state = state
w.mu.Unlock()
if !changed {
	return
}
w.uiMu.Lock()
defer w.uiMu.Unlock()
if ctx.Err() != nil {
	return
}
handleFSChange(w.ex, flag, uri)

File: internal/ide/out_of_root_watch.go [hunk 9]
```diff
@@ -0,0 +1,284 @@
...
+		}
+	case textapi.EventTypeClose:
+		w.remove(ev.URI)
+	}
+	return false
+}
+
+func (w *outOfRootWatcher) isOutOfRoot(uri workspaceapi.URI) bool {
+	// Internal tabs (memory://, terminals) have no file on disk.
+	if uri.Scheme() != "file" || !filepath.IsAbs(uri.Path()) {
```
ernestrc, line 139: This used to be `WithPath` plus `Equal`, which compared scheme *and* host against the root. It is now a hardcoded `uri.Scheme() != "file"`.

A workspace rooted on a non `file` scheme (`memory://`, `docs://`) will classify every `file://` tab as out-of-root and stat it through that scheme on every tick. It is benign today, both the baseline and the poll come back `exists: false` so nothing fires, with a debug log per tick at worst. Comparing against the root URI's own scheme instead of a literal restores the old invariant for free.

internal/ide/out_of_root_watch.go
```diff
@@ -0,0 +1,284 @@
...
+
+func (w *outOfRootWatcher) add(uri workspaceapi.URI) {
+	state := w.stat(uri)
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	if w.closed {
+		return
+	}
+	w.tabs[uri.String()] = &outOfRootTab{uri: uri, state: state}
+	if w.cancel == nil {
```
ernestrc, line 157: I don't like sort of initialization once some later event occurs. If there ever were to be a bug down the line, this leads to users not being to establish a repro case. Initialize a worker goroutine during the initializer/constructor. This way if there's a bug, or any measurable side effect, it'll be immediate, not predicated on user action.

This also introduces the nice property now that since `context.CancelFunc` functions are designed to be idempotent, you don't need to check for nil in [stopPolling](https://github.com/unstablebuild/rune/pull/79/changes#diff-0db23aeb595753dc66d0dde76c93ecd9d9c65ec3b70e098a50b53177a6536e46R181). This makes the code easier to read and simpler.

Issue #78, opened 2026-09-16
Compaction persists a length-truncated summary, and the resumed turn degenerates into a closing-tag loop


ernestrc:
## Summary

When the compaction summary stream ends with `FinishReasonLength`, `agent.Summarize`
accepts the partial output anyway. `cleanSummary` cannot match the `<summary>...</summary>`
pair because the closing tag never arrived, so the raw fragment, including the dangling
opening tag, is written verbatim into the post-compaction resume message and the original
conversation is archived and replaced. On the next turn the model is handed an
unterminated XML document cut mid-sentence, and it tries to repair it instead of resuming
the task.

## Observed failure

Seen on a long auto-compacted session. The archived dialogue contains no `compact` tool
call and its last message is a plain tool result, confirming the `AutoCompactRatio` path.

- Persisted resume message: 7022 chars, contains `<summary>`, does **not** contain
  `</summary>`, and ends mid-sentence inside an unclosed fenced code block.
- The following assistant turn opened with `(continued from summary)`, closed the code
  fence, wrote the remaining summary sections instead of doing any work, and then emitted
  `</parameter>\n</invoke>` **22 times** before finally closing with `</summary>`.

The visible symptom is a wall of stray closing tags in the chat transcript. The real
damage is that a turn is wasted continuing a document rather than executing the task, and
the compacted dialogue permanently carries a malformed prompt.

## Where

`cmd/rune-agent/agent/agent.go:1862-1882`. The stream loop only consumes `EventTextDelta`
and `EventStreamError`. `llmapi.Event.DoneData.FinishReason` is ignored, and the only
guard is:

```go
summary := cleanSummary(sb.String())
if summary == "" {
	return "", fmt.Errorf("LLM returned an empty summary")
}
```

`cleanSummary` (`agent.go:1832`) uses `reSummary = (?s)<summary>(.*?)</summary>`. With no
closing tag there is no match, so the text, opening tag included, passes through as-is.

`CompactDialogue` (`agent.go:1893`) then calls `store.ArchiveAndReplace`, so the truncated
summary replaces the real conversation.

Note that Rune already recognises this finish reason elsewhere: `agent.go:933`,
`cmd/rune-agent/extension/handler.go:1961` and `handler.go:2391`, which surfaces
`truncatedTurnHint` (`handler.go:1918`) for truncated chat turns. The compaction path has
no equivalent.

## Proposed fix

[...]

PR #80, opened 2026-09-16
Fix compaction summary to use session max-output-token budget

### LINKED ISSUE #77: Compaction summary request sends no max-output-token budget and silently falls back to the provider default
## Summary

`agent.Summarize` builds its completion request with only `Messages` set. It never
populates `llmapi.Request.MaxOutputTokens`, so the provider client applies its own
fallback, which is far smaller than the budget the session is configured with. On a long
conversation the summary is cut off mid-sentence, and the truncated fragment is then
persisted as the post-compaction resume prompt.

## Where

`cmd/rune-agent/agent/agent.go:1844`

```go
func Summarize(ctx context.Context, svc llmapi.Service, model llmapi.ModelEntry, messages []llmapi.Message) (string, error) {
	...
	summaryReq := llmapi.Request{
		Messages: append(messages, prompt),
	}
```

Compare the chat path, which does set it, at `cmd/rune-agent/agent/agent.go:802`:

```go
MaxOutputTokens: a.MaxOutputTokens(),
```

`Agent.MaxOutputTokens()` holds the session value (set by `/max_tokens`, validated against
the bound model at `cmd/rune-agent/extension/command_adapter.go:482` via
`llmarg.ValidateMaxOutputTokens`). `agent.compact` (`agent.go:1642`) does not pass it
through to `CompactDialogue`, and `CompactDialogue` (`agent.go:1893`) does not pass it to
`Summarize`.

## Impact

With `MaxOutputTokens == 0` every provider substitutes its own default:

- `internal/llm/anthropic/api.go:64-78` falls through to `params.MaxTokens = 8192`.
  `Config.AnthropicClientConfig()` in `internal/llm/config.go:160` never sets
  `anthropic.Config.MaxTokens`, so the middle branch cannot fire either.
- `internal/llm/openai/client.go:431` falls back to `config.MaxCompletionTokens`.
- `internal/llm/gemini/client.go:146` falls back to the provider config limit.

The compaction prompt asks for a 9-section summary plus an `<analysis>` scratchpad, and
reasoning tokens share the same budget. On a long session (observed: 361 messages, 133
assistant turns, 224 tool results) the response hits the cap and the stream stops
mid-sentence. The result is a resume prompt containing an unterminated `<summary>` block
cut inside a fenced code block. See #78 for what the model does with it.

## Proposed fix

Plumb the session's max-output-token value from `Agent` into `CompactDialogue` and
`Summarize`, and set it on the request:

1. Add a field to `compactOptions` (for example `WithMaxOutputTokens(int)`) or extend the
   `CompactDialogue` and `Summarize` signatures. Both are exported and have two production
   call sites: `agent.compact` (`agent.go:1642`) and the agentshell `/compact` command
   (`cmd/rune-agent/agentshell/agentshell.go:1056`).
2. In `Summarize`, set `summaryReq.MaxOutputTokens` when the value is non-zero.

**Do not clamp inside the agent against a provider-specific ceiling.** The summarize
service may be any provider. The session value is already validated against the bound
model by `llmarg.ValidateMaxOutputTokens` when it is set, and each provider client applies
its own ceiling (for example `effectiveMaxOutputTokens` at
`internal/llm/gemini/client.go:160`). Passing the value through unmodified is the correct
provider-agnostic behaviour. When it is `0` the existing provider defaults still apply.

## Tests

Per the repo bug-fix policy, start with a failing test:

- In `cmd/rune-agent/agent/agent_test.go`, assert that the request captured by the fake
  `llmapi.Service` used by the existing compaction tests carries the configured
  `MaxOutputTokens`. Table-driven: a configured value, and `0` meaning "leave unset".
- Assert the agent does not rewrite the value (no clamping), so the test stays
  provider-neutral.

## Verification

```bash
go test -race -run TestCompact ./cmd/rune-agent/agent/
make generate && make lint && make test
```

santhiprakash:
## Problem

`agent.Summarize` builds its completion request with only `Messages` set and never populates `llmapi.Request.MaxOutputTokens`. The provider client then applies its own fallback (`params.MaxTokens = 8192` for Anthropic, `config.MaxCompletionTokens` for OpenAI, the provider config limit for Gemini) — much smaller than the budget the session is configured with. On a long conversation the summary stream stops mid-sentence, and the truncated fragment is then persisted as the post-compaction resume prompt. #78 shows the model looping on the unterminated `<summary>` block.

Compare the chat path at `cmd/rune-agent/agent/agent.go:802`, which already sets `MaxOutputTokens: a.MaxOutputTokens()`.

## Triage

The session value lives on `Agent` (`SetMaxOutputTokens` / `MaxOutputTokens`, set by `/max_tokens`, validated against the bound model by `llmarg.ValidateMaxOutputTokens` at `cmd/rune-agent/extension/command_adapter.go:482`). `agent.compact` (`agent.go:1642`) does not pass it through to `CompactDialogue`, and `CompactDialogue` (`agent.go:1893`) does not pass it to `Summarize`. The chat path is the only site that wires it correctly.

## Fix

Plumb the session's max-output-token value through to the summarize request:

- Add `WithMaxOutputTokens(int)` to the existing `compactOptions` (alongside `WithCompactHooks`).
- `CompactDialogue` reads `copts.maxOutputTokens` and passes it to `Summarize`.
- `Summarize` now takes a `maxOutputTokens int` parameter and sets `summaryReq.MaxOutputTokens` when non-zero. A value of `0` leaves the provider fallback default in place, matching today's behaviour for callers that don't pass a budget.
- `agent.compact` passes `WithMaxOutputTokens(a.MaxOutputTokens())` so the session value reaches the summarize request.
- The agentshell `/compact` command keeps the prior behaviour — the compact service is resolved against a separate alias and there is no session value to pass through.

Per the issue, I did not clamp the value against a provider-specific ceiling. The session value is already validated against the bound model, and each provider client applies its own ceiling (`effectiveMaxOutputTokens` at `internal/llm/gemini/client.go:160`).

## Verification

Regression test `TestCompactDialoguePlumbsMaxOutputTokens` in `cmd/rune-agent/agent/agent_test.go` (table-driven across `WithMaxOutputTokens` set / unset / explicit zero). Sabotage-verified: reverting only the `if maxOutputTokens > 0` block in `Summarize` fails the "set option" subtest with `expected 8192, actual 0`.

```
$ go test -race -count=1 ./cmd/rune-agent/agent/
ok  	unstable.build/rune/cmd/rune-agent/agent	2.677s
```

## Notes / Risks

- `CompactDialogue` and `Summarize` are exported; `Summarize`'s new `maxOutputTokens int` parameter is a breaking signature change for any out-of-tree caller. The only internal caller is `CompactDialogue`, updated here.
- The agentshell `/compact` path is intentionally untouched. If the maintainer wants that path plumbed as well, follow-up is straightforward — pass the shell's session config through a new `WithMaxOutputTokens` call.
- DCO trailer included on the commit; `go build`/`go vet` clean for the touched package.

Fixes #77


santhiprakash:
Note: the four CI workflows (Linux / Lint / Security / macOS) are showing `action_required` — this is the standard first-time fork-contributor workflow-approval gate. A maintainer needs to click **Approve and run** on each of the four workflow runs at https://github.com/unstablebuild/rune/pull/80/checks before CI will start. Not an author-side failure.

Pre-flight verification I ran locally before pushing:

```
$ go vet ./cmd/rune-agent/agent/                  # clean
$ go build ./cmd/rune-agent/agent/                # clean
$ go test -race -count=1 ./cmd/rune-agent/agent/  # ok 2.677s
$ go test -race -count=1 -run TestCompactDialoguePlumbsMaxOutputTokens ./cmd/rune-agent/agent/  # 3/3 sub-tests pass
```

Sabotage-verified: reverting only the `if maxOutputTokens > 0` block in `Summarize` fails the new "set option" subtest with `expected 8192, actual 0`.

`make generate` and `make lint` need `mockgen` / `protoc` / `golangci-lint` which aren't in my sandbox; the repo's pre-submit check is documented as `make generate && make lint && make test`, so please let me know if any of those surface an issue I should fix.

cmd/rune-agent/agent/agent.go
```diff
@@ -1840,8 +1843,10 @@ func cleanSummary(raw string) string {
 }
 
 // Summarize sends the given messages to the LLM and asks it to produce
-// a structured summary. It returns the cleaned summary text.
-func Summarize(ctx context.Context, svc llmapi.Service, model llmapi.ModelEntry, messages []llmapi.Message) (string, error) {
+// a structured summary. It returns the cleaned summary text. When
+// maxOutputTokens is greater than zero it is set on the request so the
+// provider's smaller default cap does not truncate long summaries.
+func Summarize(ctx context.Context, svc llmapi.Service, model llmapi.ModelEntry, messages []llmapi.Message, maxOutputTokens int) (string, error) {
```
ernestrc: nit: can we wrap this function so it doesn't cross 90 characters wide?

cmd/rune-agent/agent/agent_test.go
```diff
@@ -1007,6 +1007,95 @@ func searchSubstring(s, substr string) bool {
...
+			},
+			Version: 1,
+		}
+		store.mu.Lock()
+		store.data["d"] = d
+		store.mu.Unlock()
+
+		_, _, err := CompactDialogue(
+			context.Background(), svc, llmapi.ModelEntry{}, store, d,
+			WithMaxOutputTokens(8192),
```
ernestrc, line 1032: This test is a good addition but it's insufficient. We should also ensure that we can't, by mistake, 			drop `WithMaxOutputTokens(8192)` from agent.go, and suddently regress to exactly where we were prior to this PR. 

I think you should be able to reuse the `TestAutoCompact` harness: call `ag.SetMaxOutputTokens(8192)` before `ag.Run(...)`, then assert `svc.requests[0].MaxOutputTokens == 8192` (request 0 is the summarize request).

cmd/rune-agent/agent/agent.go
```diff
@@ -1639,7 +1639,10 @@ func (a *Agent) compact(
 		summarizeSvc = a.config.CompactSvc
 	}
 
-	compactedMsgs, archivedID, err := CompactDialogue(ctx, summarizeSvc, a.config.Model, a.store, d)
+	compactedMsgs, archivedID, err := CompactDialogue(
+		ctx, summarizeSvc, a.config.Model, a.store, d,
+		WithMaxOutputTokens(a.MaxOutputTokens()),
```
ernestrc: for the overwhelming majority of runs a.MaxOutputTokens is 0, which is the default value, unless `SetMaxOutputTokens` is called, and it is only if the user's config has max_tokens (not the default) or the user sets it via `/max_tokens` inline command, or `chatmaxtokens` command, so `Summarize` [skips the assignment](https://github.com/unstablebuild/rune/pull/80/changes#diff-408497fcb6decc62e0440cac93a88cb6337388378552aa25d8b320f5171d33fdR1859) and the provider [falls back to the default i.e 8192](https://github.com/unstablebuild/rune/blob/48883dce5c888cd33315cd5ebade59b4eff73b00/internal/llm/anthropic/api.go#L77). The commit subject says "Fix compaction summary to use session max-output-token budget," but on a default install there is no session budget to use.

I suggest we either set a summarize default that's higher than 8192, either here with a max() statement, or set a default `max_tokens` higher in the default config, which would apply to all agent runs, not just summaries.

cmd/rune-agent/agent/agent.go
santhiprakash: Wrapped the signature in the latest push.

santhiprakash (commented):
Pushed an update addressing the review feedback:

- Wrapped `Summarize` so the signature stays within 90 columns.
- Added a model-capped default summary budget (`32768`, clamped to the bound
  model's documented ceiling via `llmarg.MaxOutputTokens`) that applies when
  the session has not set an explicit `max_tokens` override. This fixes the
  default-install case where the Anthropic 8192 provider fallback truncated
  long summaries.
- Added `TestAutoCompactPlumbsMaxOutputTokens` and
  `TestAutoCompactUsesDefaultMaxOutputTokens` using the `TestAutoCompact`
  harness to cover both the explicit session override and the default budget
  paths.

Local verification: `go vet`, `go build`, and `go test -race` for
`./cmd/rune-agent/agent/` all pass.

santhiprakash:
Pushed an update addressing the review feedback:

- Wrapped `Summarize` so the signature stays within 90 columns.
- Added a model-capped default summary budget (`32768`, clamped to the bound
  model's documented ceiling via `llmarg.MaxOutputTokens`) that applies when
  the session has not set an explicit `max_tokens` override. This fixes the
  default-install case where the Anthropic 8192 provider fallback truncated
  long summaries.
- Added `TestAutoCompactPlumbsMaxOutputTokens` and
  `TestAutoCompactUsesDefaultMaxOutputTokens` using the `TestAutoCompact`
  harness to cover both the explicit session override and the default budget
  paths.

Local verification: `go vet`, `go build`, and `go test -race` for
`./cmd/rune-agent/agent/` all pass.

ernestrc (approved):

santhiprakash:
Shipped in `11db287657af9ca7e26e4d9c40658ea5216d0070` — thanks for the review and merge, @ernestrc.

Issue #74, opened 2026-09-16
Tutorial step 1 doesn't respect system keyboard language


braaar:
### What happened

Hi, I have just downloaded and launched Rune as I'm writing this issue.

The very first thing the app asks me to do is to is make the text bigger if I want, by pressing `CMD +`. I promptly do that and the text becomes... smaller. 

You see, I'm on a nordic keyboard layout where `+` is next to the 0-key and `-` is next to left shift. I suspected something like that may be happening and confirmed that the zoom behaviour does indeed conform to the position of the minus and plus keys on a US QWERTY keyboard layout. 

Perhaps there is some config step or something later in the tutorial that fixes this, but I figured this was worth mentioning since it can create a bad first impression.
 

### What you expected

I expected the text to become larger

### Steps to reproduce

Have a Nordic keyboard layout, Install Rune v1.2.1, launch the tutorial, try to zoom in with `CMD +`

### Area

Configuration and key bindings

### Rune version

Rune v1.2.1

### OS and architecture

MacOS 27.0 arm64

### Logs, crash report, or screenshots

```shell

```

### Relevant configuration

```yaml

```

### Checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest released version.

braaar:
In a similar vein, I get stuck on the tutorial step for switching tabs. `<alt-]>` isn't available on my keyboard, since the `]` symbol is tucked away behind the key combination `alt-9`. I had to look up a US keyboard layout to fins out what key I should press here. Turns out it's the `¨`/`^` key!

braaar:
Cheatsheet at `<alt-/>` is also in a different spot. `alt--` for me

ernestrc:
This should have been fixed with #85.

Issue #67, opened 2026-09-14
tutorial step 12 (fullscreen) doesn't work


schmurfy:
### What happened

I tried on my Linux machine and osx and on both I am stuck here:

<img width="1183" height="176" alt="Image" src="https://github.com/user-attachments/assets/b74de7e6-3c5d-4137-bbe3-2738e515a35a" />

`alt+m` does nothing and I cannot progress, I am using `1.2.1`

### What you expected

I expected the keybind to work and the pane to be maximized to advance the tutorial

### Steps to reproduce

1. Launch rune for the first time
2. advance to step 12

### Area

Configuration and key bindings

### Rune version

Rune v1.2.1 (HEAD is v1.2.1)

### OS and architecture

macOS 15.7.9 arm64

### Shell

GNU bash, version 5.2.37(1)-release (aarch64-apple-darwin22.6.0)

### Logs, crash report, or screenshots

```shell

```

### Relevant configuration

```yaml

```

### Checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest released version.

rune-agent:
Thanks for the report. Can you run the `keydump` command, within Rune's command prompt, and paste the output here?

It should be something like this

<img width="486" height="416" alt="Image" src="https://github.com/user-attachments/assets/9ab05710-dfe0-4506-a1fb-5a35f6e5714c" />

If you don't see anything, it means that `<alt-m>` is not reaching Rune, therefore something at the OS level is using it.

schmurfy:
both keyboards are azerty, keydump shows:
```
<alt-;>
```

what is weird is that the "a" key correctly shows "a", same with "q" or "m".

but pressing `alt+a` shows `<alt-q>`, same as pressing `alt+m` shows `<alt-;>`

ernestrc:
This should have been fixed in #85

PR #84, opened 2026-09-16
Fix modifier chords to resolve keys through the keyboard layout

### LINKED ISSUE #59: Ctrl+Shift+P reads scancode instead of layout keys
### What happened

I type on Colemak. When I press Ctrl+Shift+P, nothing happens. When I press Ctrl+Shift+; I get the command prompt. On Colemak, the US ANSI physical key P prints key ; whereas physical key R maps to layout key P. 

### What you expected

Keyboard shortcuts need to respect layout

### Steps to reproduce

1. Set OS keyboard layout to Colemak 
2. Press Ctrl+Shift+P (physical keys Ctrl+Shift+R).

### Area

Configuration and key bindings

### Rune version

v1.2.1

### OS and architecture

Linux Ubuntu 26.04 amd64

### Shell

_No response_

### Logs, crash report, or screenshots

```shell

```

### Relevant configuration

```yaml

```

### Checks

- [x] I can reproduce this on the latest released version.

santhiprakash:
Closes #59

#### Problem

On a non-QWERTY layout, control/command chords fire on the physical key
position instead of the layout character. On Colemak, `Ctrl+Shift+P` does
nothing and `Ctrl+Shift+;` opens the command prompt; on Dvorak, `alt+x`
requires pressing `alt+q` (see the confirmation comment on the issue).
This is also the root cause class behind #67 and #74.

#### Root cause

`internal/term/gui/input.go` `processEvents` used `ev.Key` directly.
Ebiten key constants name the US QWERTY position of a key, so
`keyToBaseAndShift` synthesized the QWERTY character for the position —
physical `R` produced `<c-s-r>` on Colemak, never matching the `<c-s-p>`
binding (which parses to `{Mod: ModCtrl, Ch: 'P'}`). The same physical
`Key` also keyed the `gui.key_mapping` remap table, so configured remaps
matched positions rather than characters.

#### Fix

Resolve character-producing keys through the platform key name
(`ebiten.KeyName`, i.e. `glfwGetKeyName`) before the `gui.key_mapping`
lookup and character synthesis. `KeyName` reports the unshifted
character the position produces under the active layout; `runeKey` maps
that character back to the key whose US name produces it, so the
existing `keyToBaseAndShift`/`resolveCharKey` path emits the layout
character. Colemak `Ctrl+Shift+physical-R` now emits
`{Mod: ModCtrl, Ch: 'P'}` and opens the command palette.

Details:

- Numpad keys and non-printable keys keep their physical identity (the
  platform key-name lookup does not cover them).
- A position whose layout character has no US-named key (non-Latin
  layouts, dead keys, multi-character names) falls back to the reported
  physical key — same as today, no regression.
- Plain (unmodified) typing is untouched: characters still come from the
  platform's own text commit.
- `gui.key_mapping` remaps now match the layout key as well, since the
  lookup uses the resolved key — a configured `<c-p>` fires on the key
  that types `p`.

#### Verification

New table-driven regression test `TestKeyDispatchLayoutAware` in
`internal/term/gui/input_test.go` covers the reported scenario plus
siblings: Colemak `ctrl+shift+p`, the physical-`P` position resolving to
`;`, Dvorak `alt+x`, empty/unmapped key-name fallbacks, plain-text
commit passthrough, `key_mapping` matching the layout key, and numpad
physical identity.

- `xvfb-run -a go test -race -run TestKeyDispatchLayoutAware ./internal/term/gui/` — 8/8 pass
- Sabotage-checked: reverting the dispatch line to `ev.Key` fails the Colemak, Dvorak, and remap subtests (emits `R`/`q` instead of `P`/`x`)
- `xvfb-run -a go test -race -count=1 ./internal/term/gui/ ./internal/ide/` — ok
- `go build ./...` clean; `go vet ./internal/term/gui/` clean (`./internal/ide/` shows only pre-existing unkeyed-field warnings on `main`)

#### Notes

The platform name lookup reports the unshifted character only, so a
shifted chord keeps the US shift pairing of the resolved key — which is
also how `term.ParseKey` folds shift into binding specs, so binding
identity stays consistent.


santhiprakash:
The four GHA workflows show `action_required` (first-time-contributor gate) — they need a maintainer approval click to run. Local verification in the meantime:

- `xvfb-run -a go test -race -run TestKeyDispatchLayoutAware ./internal/term/gui/` — 8/8 pass (new regression test; fails on the pre-fix code)
- `xvfb-run -a go test -race -count=1 ./internal/term/gui/ ./internal/ide/` — ok
- `go build ./...` clean; `go vet ./internal/term/gui/` clean

ernestrc:
Thanks so much for taking this! I've also been working on a fix, which goes one step further than this PR. See #67 and #74. 

The problem is that `ebiten.KeyName(key)` can get us so far. It translates the key as such: character -> runeKey reverse map -> back to a US `ebiten.Key` -> `keyToBaseAndShift` which re-derives the character from a US keyboard layout table. That round trip through the US key enum is the crux. It's a lossy channel: only characters that are a base character of some US key can survive it. This has the following implications:

1. Non-US characters fall back to the physical key: still wrong. Nordic ¨ (US ]), ø/æ, German ü ö ä, and all of Cyrillic/Greek/Hebrew have no US-named key, so layoutKey returns the physical key and you get the original bug. Your test documents this as acceptable ("unmapped key name falls back to the physical key", with я), but we should try to get to the bottom of it, like any other desktop application does.

2. AZERTY's top row breaks. &, é, " are shift characters in the US table, never bases, so runeKey misses them -> fallback to Digit1.. etc. So #67 letters work but the digit row doesn't.

 Fortunately, we can rely on glfw's platform level code to actually pass the actual code points to the event, so we don't have to guess or convert anything. See https://github.com/unstablebuild/rune/pull/85

santhiprakash:
Agreed — taking the codepoints straight from the platform via `ev.Char`/`ev.ShiftChar` is the right fix. The `KeyName` → US-key-enum round-trip here can't represent non-US characters or AZERTY's shifted digit row, exactly the two gaps you listed. Closing in favor of #85.

PR #88, opened 2026-09-17
Add agent skills reload command with differential reporting

### LINKED ISSUE #76: Rune Agent: no way to explicitly reload skills and see what the registry picked up
### Problem

Rune Agent's skill registry re-scans its tracked skill directories implicitly:
once per agent turn, and once on every `agent` command. That keeps skills
fresh, but it gives the user no way to trigger a re-scan deliberately and no
feedback about what the re-scan found.

So when you are authoring a skill, the loop is guesswork. You edit a
`SKILL.md`, then run some unrelated `agent` subcommand to nudge the implicit
reload, then run `agent skills list` and diff the output by eye to check
whether your change was picked up. If the skill fails to parse, it is written
to the log at warn level and skipped, so in the REPL it simply does not appear.
The same goes for a skill saved to a directory that is not tracked. There is no
way to distinguish "not reloaded yet" from "reloaded and rejected".

### Proposed behavior

Add an `agent skills reload` subcommand that performs the re-scan explicitly
and reports the result.

The point of the command is the feedback, not the reload. It should tell the
user what actually happened to the registry: which skills are now loaded, and
ideally what changed relative to the previous state (added, updated, dropped).
A skill whose `SKILL.md` could not be parsed should be surfaced as a problem in
the command's output rather than only in the log.

Scope should stay on the directories the registry already tracks. Picking up
newly configured directories is what `agent skills add-dir` is for, and that
distinction should remain visible in the command's output, so a user who saved
a skill outside a tracked directory can tell why it did not appear.

The subcommand should sit alongside the existing `skills` subcommands: listed
in the command manual, and offered by tab completion.

The implicit per-turn and per-command reload behavior should not change.

### Area

Rune Agent

### Additional context

The skills registry and its reload logic live in
`cmd/rune-agent/agent/skills/registry.go`. The implicit reload call sites are
in `cmd/rune-agent/agent/agent.go` (per agent turn) and
`cmd/rune-agent/agentshell/agentshell.go` (per shell command).

The `skills` subcommand surface is all in
`cmd/rune-agent/agentshell/agentshell.go`: dispatch, the command manual entry,
the existing `list`, `show`, `list-dirs`, `add-dir` and `remove-dir`
implementations, and tab completion. `addSkillDir` there is the closest
existing example of a skills command that reports what the registry
discovered.

`cmd/rune-agent/agentshell/dream_test.go` shows the pattern for testing an
agent shell subcommand with in-process stubs.

Note that the current reload entry point returns no value, so reporting what
changed will likely need a richer result from the registry rather than just a
post-reload listing. Shadowed skills and over-long descriptions already reach
the user through the registry's notification hook, but parse failures do not.

Spandan-Mishra:
## Issue

Closes #76 

## Feature & Changes Added

This PR adds an explicit `agent skills reload` REPL command to re-scan all tracked skill directories on demand and provide immediate visibility into skill discovery, differential changes, and frontmatter validation errors.

### Changes:
- **`cmd/rune-agent/agent/skills/loader.go`**: Updated `loadDir` to collect and return parse/validation errors (`SkillError`) instead of silently ignoring failed skills.
- **`cmd/rune-agent/agent/skills/registry.go`**: 
  - Added `SkillError` and `ReloadResult` types.
  - Enhanced `Reload()` to compute differential results (`Added`, `Updated`, `Dropped`), surface collection errors (`Errors`), and preserve built-in skills (`explore`, `plan`) across re-scans.
  - Added `skillEqual()` helper to detect content, metadata, and configuration changes.
- **`cmd/rune-agent/agentshell/agentshell.go`**:
  - Registered `reload` in `commandManual` and `completeSkills` tab completion.
  - Implemented `reloadSkills()` markdown formatter displaying tracked directories, diff breakdowns, syntax/validation errors, and active skills.
  - Guarded pre-dispatch implicit reload in `HandleCommand` so `skills reload` can accurately compute diffs against pre-reload state.
- **Tests**:
  - Added differential reload and error capture tests in `registry_test.go`.
  - Added end-to-end command dispatch, diff rendering, and completion tests in `skills_test.go`.

---

### 1. Reload with No Changes
screenshot-2026-09-17_18-16-18

### 2. Addition of New Skill
screenshot-2026-09-17_18-13-20

### 3. Update Skill
screenshot-2026-09-17_18-15-58

### 4. Dropping Skill
screenshot-2026-09-17_18-16-41

### 5. Broken Skill Added (Syntax / Parse Error)
screenshot-2026-09-17_18-12-48

---

## Test Verification

- `go test -count=1 -race ./cmd/rune-agent/agent/skills/...` (PASS)
- `go test -count=1 -race ./cmd/rune-agent/agentshell/...` (PASS)


cmd/rune-agent/agentshell/agentshell.go
```diff
@@ -423,17 +424,20 @@ func (s *shell) handleCommand(
...
 		if len(cmd.Args) == 0 {
 			return s.Help(ctx, nil)
 		}
 		cmd = repl.Command{Name: cmd.Args[0], Args: cmd.Args[1:]}
 	}
 
+	// Re-scan skill directories so out-of-band changes are picked up,
+	// mirroring the agent loop's per-turn Reload in agent.go.
+	// Skip implicit reload for "skills reload" so the handler can compute the diff.
+	if s.skillRegistry != nil && !(cmd.Name == "skills" && len(cmd.Args) > 0 && cmd.Args[0] == "reload") {
```
ernestrc: skillRegistry [can't be nil](https://github.com/unstablebuild/rune/blob/b24b9d6e7344a7e6d24dca00a2116f50cb3585ef/cmd/rune-agent/agentshell/agentshell.go#L142). Same for line 1260, 1503, 1517

ernestrc: The `!(cmd.Name == "skills" && ...Args[0] == "reload")` guard only defends against the one reload immediately preceding this command. In a live session the agent loop reloads continuously, so `skills reload` will almost always print "No skill changes detected."

The registry's live state is the wrong diff baseline. `Reload()` should diff against a snapshot owned by the *reporting* caller — e.g. a `lastReported` snapshot the shell updates only when it actually renders a report, or an explicit `ReloadSince(prev map[string]Skill)`. Whatever the shape, the implicit per-turn reloads must not mutate the reporting baseline.

cmd/rune-agent/agent/skills/loader.go
```diff
@@ -24,16 +24,17 @@ import (
...
-// and skips failures.
-func (r *SkillRegistry) loadDir(dir string) []Skill {
+// registry's FileSystem. Returns all successfully parsed skills and
+// any parsing/validation errors encountered.
+func (r *SkillRegistry) loadDir(dir string) ([]Skill, []SkillError) {
 	entries, err := r.fs.ReadDir(dir)
 	if err != nil {
 		slog.Debug("skills: cannot read directory", "dir", dir, "error", err)
-		return nil
+		return nil, nil
```
ernestrc: `reload` is a command whose entire purpose is reporting what happened: a tracked directory that was deleted, renamed, or made unreadable produces "No skill changes detected" (or an unexplained Dropped list) with a Debug-level log the user never sees. This is the most likely real-world reason a user reaches for skills reload. Surface it as a SkillError with the directory path.

cmd/rune-agent/agent/skills/registry.go
```diff
@@ -190,6 +226,55 @@ func (r *SkillRegistry) Reload() {
...
+		a.Description == b.Description &&
+		a.Body == b.Body &&
+		a.Dir == b.Dir &&
+		a.License == b.License &&
+		a.Compatibility == b.Compatibility &&
+		a.AllowedTools == b.AllowedTools &&
+		a.Type == b.Type &&
+		a.Model == b.Model &&
+		a.ParentContext == b.ParentContext &&
+		maps.Equal(a.Metadata, b.Metadata)
```
ernestrc: Hand-enumerating 10 of Skill's 11 fields means any field added to Skill is silently excluded from Updated detection, with no compile error. Since Metadata is the only non-comparable field, a.Metadata/b.Metadata cleared/then reset + struct compare is safer for an equality.

cmd/rune-agent/agent/skills/registry.go
```diff
@@ -130,7 +157,7 @@ func (r *SkillRegistry) AddDir(dir string) ([]Skill, error) {
 		return nil, fmt.Errorf("directory already tracked: %s", abs)
 	}
 
-	loaded := r.loadDir(abs)
+	loaded, _ := r.loadDir(abs)
```
ernestrc: This is a great opportunity to fix this path and actually return an error so callers of AddDir benefit from it, and we don't silently ignore it.

cmd/rune-agent/agent/skills/registry.go
```diff
@@ -57,7 +79,8 @@ func NewRegistry(fs workspaceapi.FileSystem, cwd workspaceapi.URI, dirs []string
 			continue
 		}
 		r.dirs = append(r.dirs, abs)
-		for _, s := range r.loadDir(abs) {
+		loaded, _ := r.loadDir(abs)
```
ernestrc: Same here. `NewRegistry` and `AddDir` should now surface these errors. For `AddDir` this is a
visible gap: [skills add-dir <dir>](https://github.com/unstablebuild/rune/blob/b24b9d6e7344a7e6d24dca00a2116f50cb3585ef/cmd/rune-agent/agentshell/agentshell.go#L1378) is when a user finds out their SKILL.md is malformed, so we should surface it.

ernestrc (commented):
We should add the new command in the docs as well: https://github.com/unstablebuild/rune/blob/a964d798d8584e1f26098d87f5d8123bcd780bb6/cmd/rune/docs/docs/agent/skills.md#L39

cmd/rune-agent/agent/skills/loader.go
Spandan-Mishra: yeah, this makes sense

cmd/rune-agent/agent/skills/registry.go
Spandan-Mishra: ok, working on it

Issue #86, opened 2026-09-17
Add telemetry consent prompt to the bootstrap wizard


ernestrc:
## Summary

Add a third prompt to the first-run bootstrap wizard asking the user for permission to send telemetry to Rune. Today telemetry is on by default and the user is never asked; the only opt-out is editing `~/.rune/config.yaml` after the fact.

The prompt should explain that the data — in particular the per-language file-open counts — is what the Rune Authors use to decide which language support to build next, and should show a concrete example of the payload that is actually posted, rendered as a fenced code block.

**Scope:** new installs only. Existing installs never re-enter bootstrap, so they keep telemetry on by default and opt out via config as today. No startup nag, no storage-partition "has been asked" flag.

## Current state

- Bootstrap wizard: `cmd/rune/bootstrap_handler.go:868` `openBootstrapFlow()` -> `openWelcomePrompt` (`:882`) -> `openVimPrompt` (`:904`) -> `performSwap()` (`:533`) -> `writePresetConfig()` (`:779`) writes `~/.rune/config.yaml`.
- `guardedPromptChain` (`:936`) re-opens any prompt dismissed without a selection, so the wizard cannot be escaped.
- Prompt messages are already rendered as markdown (`internal/browser/component.go:1058`), including fenced code blocks (`internal/component/markdown/code.go`).
- Telemetry exists and defaults on: `telemetry.enabled` in `cmd/rune/rune.star:583`, read by `ide.TelemetryEnabled` (`internal/ide/config.go:2645`), consumed at `cmd/rune/main.go:894`.
- Payload shapes: `telemetryUsagePayload` / `telemetrySystemPayload` at `cmd/rune/ide/apiclient/telemetry.go:366-392`.
- The presets (`preset_modal.yaml`, `preset_emacs.yaml`, `preset_standard_*.yaml`) contain no `telemetry:` section today.

### 1. Third prompt in the chain

New `openTelemetryPrompt`, inserted between key bindings and the swap:

```
welcome -> key bindings -> telemetry -> performSwap
```

`openVimPrompt`'s `onSelect` stores `b.chosenEditor` and calls `b.openTelemetryPrompt()` instead of scheduling `performSwap`. The telemetry prompt's `onSelect` stores `b.telemetryEnabled` and schedules `performSwap`. Placing it last means `writePresetConfig` already knows the answer when it writes the file.

[...]

PR #90, opened 2026-09-17
Add telemetry consent prompt to bootstrap wizard

Spandan-Mishra:
### Summary
Addresses #86 by introducing an explicit telemetry consent prompt to the first-run bootstrap wizard. Previously, telemetry was enabled by default on fresh installs without an interactive prompt. Users can now choose whether to share anonymous usage reports during initial onboarding.

Telemetry consent prompt:
screenshot-2026-09-18_01-50-36

### Implementation Details
1. **Payload Example Generator without Drift (`cmd/rune/ide/apiclient/telemetry.go`)**:
   - Added `ExampleUsagePayloadJSON()` returning a formatted JSON representation of a representative `telemetryUsagePayload`.
   - Added `TestExampleUsagePayloadJSONFieldCoverage` using reflection to ensure 100% field coverage between the struct and the example output.

2. **Preset Generation & Config Migration (`cmd/rune/presets.go`, `cmd/rune/config_migrate.go`)**:
   - Updated `renderPreset(editor string, telemetry bool)` to append the explicit `telemetry:` block with documentation comments to `config.yaml`.
   - Updated config migration callers and unit tests.

3. **Bootstrap Wizard Wiring (`cmd/rune/bootstrap_handler.go`)**:
   - Added `bootstrapPrompter` interface to decouple headless prompt testing from the GUI engine.
   - Chained `openVimPrompt` -> `openTelemetryPrompt` -> `performSwap`.
   - Options: `" sure "` (bound to `y`) and `" no thanks "` (bound to `n`).
   - Protected prompt with `promptGuard()` so closing without a choice (e.g. `Esc`) safely re-prompts the user.
   - Injected consent into `performSwap`'s `rootCfg` ensuring the live API client on the first session respects the decision immediately.

4. **Documentation (`cmd/rune/docs/docs/learn/telemetry.md`)**:
   - Updated the documentation to note that fresh installs prompt for consent during onboarding and write the selection to `config.yaml`.

### Verification
- **Automated Tests**:
  - `cmd/rune/bootstrap_test.go`: Verified prompt progression, choice mapping, config persistence round-trip, and Esc dismissal re-opening.
  - `cmd/rune/ide/apiclient/telemetry_test.go`: Verified payload struct reflection coverage.
  - Full race verification: `go test -race -count=1 ./cmd/rune ./cmd/rune/ide/apiclient` (all pass).
- **Manual Verification**:
  - Created fresh test directories to test out both `y` and `n` and verified their corresponding results in the respective config.yaml of the directories

when `sure` is selected:
screenshot-2026-09-18_02-00-38

when `no thanks` is selected:
screenshot-2026-09-18_02-00-21


cmd/rune/docs/docs/learn/telemetry.md
```diff
@@ -10,8 +10,10 @@ features and languages people actually use, and which platforms and
 versions need support. This page lists every field that leaves your
 machine, and how to stop all of it with one setting.
 
-Telemetry is on by default. Everything below is sent over HTTPS to
-`api.rune.build`.
+New installs are prompted to choose whether to enable telemetry during the
+first-run bootstrap wizard, and their choice is recorded in `config.yaml`.
+Existing configurations without an explicit setting default to enabled.
+Everything below is sent over HTTPS to `api.rune.build`.
```
ernestrc, line 16: IMO we should show a sample payload here as well(on top of the table, which serves as a field + explanation).

cmd/rune/bootstrap_handler.go
```diff
@@ -865,6 +852,17 @@ const (
 	optWelcomeGo = " Let's go "
 )
 
+const (
+	optTelemetryYes = " sure "
```
ernestrc: How about "yes", "no thanks", which is a bit more neutral?

cmd/rune/bootstrap_handler.go
```diff
@@ -911,20 +924,47 @@ func (b *bootstrapHandler) openVimPrompt() {
...
+		"Rune reports a small amount of anonymous usage data. The most useful " +
+		"signal is which languages people actually edit -- it is how we decide " +
+		"which language support to build next.\n\n" +
+		"We never send file names, paths, file contents, terminal output, or " +
+		"anything you type. Here is a complete report:\n\n" +
+		"```json\n" +
+		apiclient.ExampleUsagePayloadJSON() + "\n" +
+		"```\n\n" +
+		"You can change this any time in ~/.rune/config.yaml, and the full list " +
+		"of fields is on the Telemetry page in the docs."
```
ernestrc: We should add a link to the telemetry docs; the markdown handler, if configured correctly, allows for opening a link with the browser upon clicking.

`[Telemetry](https://docs.rune.build/learn/telemetry/)`

cmd/rune/bootstrap_handler.go
```diff
@@ -911,20 +924,47 @@ func (b *bootstrapHandler) openVimPrompt() {
...
+	)
+}
+
+func (b *bootstrapHandler) openTelemetryPrompt() {
+	msg := "## Help us pick what to build next\n\n" +
+		"Rune reports a small amount of anonymous usage data. The most useful " +
+		"signal is which languages people actually edit -- it is how we decide " +
+		"which language support to build next.\n\n" +
+		"We never send file names, paths, file contents, terminal output, or " +
+		"anything you type. Here is a complete report:\n\n" +
```
ernestrc: Calling this a “complete report” is misleading. We’re only showing the usage payload here, but there’s also a system payload where we collect things like OS, architecture, kernel version, etc, and it's logged once at startup.

We probably can’t fit both payloads in this prompt, so I think we should either change the wording to make it clear that there are additional payloads, or not show a payload here at all. In that case, we should strongly emphasize that users can check the docs for a complete report of what’s collected. While we’re at it, the payload shown in the docs should be an actual real-world payload rather than one with made-up values. I think that would give users more confidence in what we’re collecting.

cmd/rune/bootstrap_handler.go
```diff
@@ -879,16 +877,31 @@ func metaKeySymbol() string {
...
+	bindings []term.KeyComb,
+	h sdkhandler.PromptHandler,
+) browser.Window {
+	if b.prompter != nil {
+		return b.prompter.Prompt(message, options, bindings, h)
+	}
+	if b.preIDE != nil {
+		return b.preIDE.Prompt(message, options, bindings, h)
+	}
+	return nil
```
ernestrc: `bh.prompter = preIDE` is set at bootstrap_handler.go:155 before `openBootstrapFlow()`, so `b.prompter` is never nil, on any prompt path. The `b.preIDE` branch is dead and return nil would silently
drop the prompt, leaving the wizard with no way forward. Collapse to return `b.prompter.Prompt(message, options, bindings, h)`. Also `performSwap` nils `b.preIDE`
 but leaves `b.prompter` pointing at the closed pre-IDE; swap that one too to the new IDE instance.

cmd/rune/bootstrap_handler.go
```diff
@@ -911,20 +924,47 @@ func (b *bootstrapHandler) openVimPrompt() {
...
+			}),
+			guard.onClose(b.openVimPrompt),
+		),
+	)
+}
+
+func (b *bootstrapHandler) openTelemetryPrompt() {
+	msg := "## Help us pick what to build next\n\n" +
+		"Rune reports a small amount of anonymous usage data. The most useful " +
+		"signal is which languages people actually edit -- it is how we decide " +
```
ernestrc: nit: replace "-- it" with "and it" or "which is".

cmd/rune/bootstrap_handler.go
Spandan-Mishra: yes this seems more neutral

cmd/rune/bootstrap_handler.go
Spandan-Mishra: replacing it with 'and it'

cmd/rune/bootstrap_handler.go
Spandan-Mishra: i think in that case not showing a payload here would be better. and as for the docs, yes we should update it to show the complete report of the things that are being collected, that would be a better approach to this.

cmd/rune/docs/docs/learn/telemetry.md
Spandan-Mishra: i have added the samples here, do let me know if there are any changes required here @ernestrc

cmd/rune/ide/apiclient/telemetry.go
```diff
@@ -390,3 +390,28 @@ type telemetrySystemPayload struct {
 	SystemRelease      string
 	SystemVersion      string
 }
+
+// ExampleUsagePayloadJSON returns a formatted JSON representation of a
+// representative usage telemetry payload for display in onboarding and documentation.
+func ExampleUsagePayloadJSON() string {
```
ernestrc: this is unused now, we can remove it

cmd/rune/docs/docs/learn/telemetry.md
```diff
@@ -21,6 +23,22 @@ There are exactly two reports.
 
 A system report is sent when Rune starts:
 
+Example startup report:
+
+```json
+{
+  "Type": "ClientStartup",
```
ernestrc: this is made up. IT's not `ClientStartup`, It's `ClientSystem`: https://github.com/unstablebuild/rune/blob/66f67dbd600c34d5b13161dae76550c6e544f082/cmd/rune/ide/apiclient/telemetry.go#L222

cmd/rune/docs/docs/learn/telemetry.md
```diff
@@ -21,6 +23,22 @@ There are exactly two reports.
...
+  "SID": "s_f7a28b1e4c90",
+  "EditorMode": "standard",
+  "OS": "linux",
+  "Arch": "amd64",
+  "NumCPU": 8,
```
ernestrc: this is not collected or sent at all. Can you double check all of these fields. We should show **exactly** what we're collecting, otherwise what's the point of showing a payload that's not real?

cmd/rune/bootstrap_handler.go
```diff
@@ -913,20 +942,44 @@ func (b *bootstrapHandler) openVimPrompt() {
...
+
+func (b *bootstrapHandler) openTelemetryPrompt() {
+	msg := "## Help us pick what to build next\n\n" +
+		"Rune reports a small amount of anonymous usage data. The most useful " +
+		"signal is which languages people actually edit, and it is how we decide " +
+		"which language support to build next.\n\n" +
+		"We never send file names, paths, file contents, terminal output, or " +
+		"anything you type.\n\n" +
+		"You can change this any time in ~/.rune/config.yaml. A complete report " +
+		"and sample payloads are on the [Telemetry](https://docs.rune.build/learn/telemetry/) page in the docs."
```
ernestrc: this markdown link is not clickable, so it's not a great experience for the user if they try to click it and it doesn't lead anywhere. Drop the link for now, I don't want to refactor how all of the prompts are installed; currently [they're not wrapped in a `markdown.Handler`, only a `markdown.Component`](https://github.com/unstablebuild/rune/blob/66f67dbd600c34d5b13161dae76550c6e544f082/internal/browser/component.go#L1061), so there's no mouse event handling.

cmd/rune/bootstrap_test.go
```diff
@@ -52,22 +75,29 @@ func TestOptionToChoiceMapping(t *testing.T) {
...
 
-	deprecated, err := renderPreset(editorModeless)
+	deprecated, err := renderPreset(editorModeless, true)
 	require.NoError(t, err)
 	require.Equal(t, std, deprecated,
 		"the deprecated modeless alias must resolve to the standard preset")
 
-	ema, err := renderPreset(editorEmacs)
+	ema, err := renderPreset(editorEmacs, false)
+	require.NoError(t, err)
```
ernestrc, line 98: this is duplicated. remove

cmd/rune/bootstrap_handler.go
```diff
@@ -913,20 +942,44 @@ func (b *bootstrapHandler) openVimPrompt() {
...
+}
+
+func (b *bootstrapHandler) openTelemetryPrompt() {
+	msg := "## Help us pick what to build next\n\n" +
+		"Rune reports a small amount of anonymous usage data. The most useful " +
+		"signal is which languages people actually edit, and it is how we decide " +
+		"which language support to build next.\n\n" +
+		"We never send file names, paths, file contents, terminal output, or " +
+		"anything you type.\n\n" +
+		"You can change this any time in ~/.rune/config.yaml. A complete report " +
```
ernestrc: "any time in ~/.rune/config.yaml" -> "any time in your config". the config data directory is fixed, and this could easily become stale if we move the configuration to ~/.config or something like that.

cmd/rune/presets.go
```diff
@@ -28,18 +28,21 @@ var presetModalYAML string
...
-		return presetStandardYAML, nil
+		body = presetStandardYAML
 	case editorEmacs:
-		return presetEmacsYAML, nil
+		body = presetEmacsYAML
+	default:
+		return "", fmt.Errorf("unknown editor choice: %q", editor)
 	}
-	return "", fmt.Errorf("unknown editor choice: %q", editor)
+	return fmt.Sprintf("%s\ntelemetry:\n  # Report anonymous usage and system information. See the Telemetry\n  # page in the Rune docs for the full list of what is reported.\n  enabled: %t\n", body, telemetry), nil
```
ernestrc: nit: wrap this line properly at ~90 characters.

cmd/rune/bootstrap_handler.go
```diff
@@ -570,6 +578,7 @@ func (b *bootstrapHandler) performSwap() error {
 			closeErr = fmt.Errorf("close pre-config ide: %w", cerr)
 		}
 		b.preIDE = nil
+		b.prompter = realIDE
```
ernestrc: nit: no prompt is open after the swap, so this is unnecessary.

cmd/rune/docs/docs/learn/telemetry.md
Spandan-Mishra: my bad, i'll check the fields and update the correct ones

cmd/rune/bootstrap_handler.go
Spandan-Mishra: ok i'll drop this. 
does this need a change though? like some way to make the links clickable, in a separate issue/PR.

cmd/rune/docs/docs/learn/telemetry.md
Spandan-Mishra: i ran an instance locally and captured the live payloads, so now the fields, keys and values in the docs are the actual values from my system. hope this is correct now!

ernestrc (approved):

PR #93, opened 2026-09-17
Add Nix flake support for NixOS packaging and development

cstrachan88:
Add flake.nix, flake.lock, .envrc, and update README.md to support building Rune
natively on NixOS and providing a hermetic Nix development shell.

### What's Included

* **`flake.nix`**: Package derivation (`rune-ide`) and development shell (`devShells.default`).
* **`flake.lock`**: Pins `nixpkgs-unstable` for reproducible builds.
* **`.envrc` & `.gitignore`**: Adds `direnv` integration (`use flake`) and ignores `.direnv/` and `result`.
* **`README.md`**: Documents Nix run, build, profile install, NixOS/Home Manager configuration, and `nix develop` instructions.

### Key Packaging Details

1. **Package Name (`rune-ide`)**:
   Sets `pname = "rune-ide"` to clearly reflect Rune's role as a full-featured IDE + multiplexer, and avoid potential package name collisions when packaged in Nixpkgs.

2. **Tree-sitter C Sources Workaround (`modPostBuild`)**:
   Rune relies on a fork of `go-tree-sitter` that includes C source files in `src/` and `include/`. Standard `go mod vendor` strips non-Go files; `modPostBuild` copies these C sources back into `vendor/github.com/tree-sitter/go-tree-sitter` prior to build.

3. **Ebitengine Runtime OpenGL Loading & Desktop Entry**:
   Ebitengine dynamically loads `libGL.so` by filename at runtime via `dlopen`. On NixOS, unversioned `.so` files are not placed in standard global dynamic linker search paths. `postInstall` uses `wrapProgram` to prefix `LD_LIBRARY_PATH` with `${pkgs.libGL}/lib`, avoiding any source patches to vendored Ebitengine code. It also installs `deploy/rune-linux/rune.desktop` and hicolor icons for application menu launchers (fuzzel, rofi).

4. **Development Shell**:
   Includes required build dependencies (`alsa-lib`, `libGL`, `libX11`, `wayland`, `pkg-config`, `xvfb-run`), Go toolchain, and sets `LD_LIBRARY_PATH` in `shellHook` so running `./bin/rune` or `go run ./cmd/rune` works seamlessly inside `nix develop`.

5. **Version Handling**:
   `version = "1.2.1"` is currently defined inside `flake.nix` to keep this PR self-contained without changing existing release scripts or Makefiles. If you prefer using a root `VERSION` file as a single source of truth across the Makefile and `flake.nix`, I am happy to update the PR.


ernestrc (approved):

ernestrc (changes requested):
Thanks so much for putting this together. Two findings here are genuinely useful and
neither was on our radar: the `modPostBuild` workaround for the C sources that
`go mod vendor` strips out of our `go-tree-sitter` fork, and the fact that a
`.desktop` entry plus hicolor icons are needed for launcher integration.

Some context that will shape how we land this. I have #105 open, which adds a
`dist/` tree with Debian, Arch, and Homebrew packaging: Dockerized builds behind
`make pkg-deb` and `make pkg-arch`, a CI workflow, and a `.SRCINFO` drift check.
The intent is that everything distribution related lives under `dist/`, so I
would like to fold Nix into the same structure rather than have it sit at the
repo root on its own.

That said, this PR is really two things, and I would like to treat them
separately.

## The dev shell

`.envrc`, `devShells.default`, `nix develop`, and the "development" README
section. No objections, this is a nice quality of life addition and I would take
it as is, modulo two things:

1. It is pinned to `x86_64-linux`, which means an aarch64-linux contributor
   cannot `nix develop` at all. Worth a `forAllSystems` or `flake-utils` wrapper.
2. `modPostBuild` hardcodes `go-tree-sitter@v0.25.0-ub.1`, which will break
   silently on the next dependency bump. Deriving it from `go.mod`, or at least
   failing loudly, would save someone a confusing afternoon.

## The package

`packages.default`, `apps.default`, and the README section advertising
`nix run`, `nix profile install`, and `configuration.nix`. This one crosses into
being a published install channel, the same category as our Homebrew cask, so it
needs to meet the same contract as the other packages in #105. Concretely:

- **`-X unstable.build/rune/internal/debug.OSPackaged=true`.** #105 adds a build
  time switch that disables the in-product updater when a package manager owns
  the install. `upgrade` then prints "This build was distributed by an OS package
  manager, so it has disabled auto-updates. Check your distribution's package
  manager." This matters more on Nix than anywhere else, since the updater would
  be trying to write into a read-only `/nix/store` path.

- **The `rune.app/bin` install layout.** `linuxAppDir` in `cmd/rune/main.go` only
  bootstraps the bundled zsh profile when `/proc/self/exe` resolves into
  `.../rune.app/bin/`. Installing to `$out/bin/rune` silently drops that
  integration. The deb and Arch packages install to `usr/lib/rune.app/bin/rune`
  with a symlink at `usr/bin/rune`.

- **Version derived from the tag** rather than a hardcoded `1.2.1`. You flagged
  this yourself, and yes, a single source of truth is the right call.

- **Library coverage.** This is the one I would look at hardest. Since f568e55
  ("Add --headless flag so Rune can join network without graphical libraries")
  the ebiten fork resolves Xlib through `dlopen`, so `libc.so.6` and `libm.so.6`
  are the only `DT_NEEDED` entries left in the binary. Everything graphical, X11
  and xkbcommon and wayland included, is now loaded at window-open time, not just
  `libGL`. That is why the deb ships those as `Recommends` and Arch as
  `optdepends`: `--tui` and `--headless` must work on a host with no graphical
  stack at all.

  The Nix specific consequence is that the default `fixupPhase` runs
  `patchelf --shrink-rpath`, which drops RUNPATH entries for libraries that are
  not in `NEEDED`. So the store paths for X11 and friends get stripped even
  though they are listed in `buildInputs`, which I believe is the real reason
  `libGL` needed the wrapper in the first place. I suspect the current
  `wrapProgram` only gets you a working window under Wayland, and X11 will fail
  at `dlopen`. Worth testing on both.

- **`--prefix LD_LIBRARY_PATH` leaking into child processes.** Rune is a
  multiplexer. It spawns shells, language servers, and debuggers, all of which
  inherit that variable. Prefixing nixpkgs' Mesa onto their environment is a real
  hazard, and hardcoding it also breaks users on proprietary drivers who need
  `/run/opengl-driver/lib`.

## Suggestion

Rebase against my branch so you can use the ldflag and and then add `dist/nix/package.nix` holding
the actual derivation, with the root `flake.nix` importing it via `callPackage`
so the two cannot drift. `flake.nix` has to stay at the repo root for
`nix run github:unstablebuild/rune` to resolve, but the derivation itself belongs
in `dist/` with everything else.

Longer term the real target is a nixpkgs submission so the derivation already living in `dist/nix/` makes that submission a copy
rather than a rewrite.

Happy to do the `dist/nix/` integration myself on top of #105 if you would rather
not carry it. Your `modPostBuild` work would come across intact.

ernestrc:
Also, please sign your commits via `-s` so you pass the DCO check.

cstrachan88:
Excellent. Great notes and feedback. You've given me a clear path forward and I should be able to address most, if not all of this tomorrow. Traveling at the moment.

My original goal was just to get a flake working so I could test and tinker, but I figured it was a good opportunity to clean it up and submit as a PR. As such, I only focused on getting it working with linux x64. Full platform support, both for building from the flake and for development, was going to be my next step.

I'll rework the package around the `dist/nix/package.nix` structure and address the packaged build requirements you called out. Longer term, it would be nice to get Rune into the nix package repository. The `rune` package name was taken so I went with `rune-ide` for the flake package name. Happy to change it to whatever you prefer.

And yes, I went through the contribution guide to make sure everything adhered, but totally forgot the commit signing. Doh. Will fix that as well.

-Courtney

cstrachan88:
Oh and how do you want to handle the version string? From my tests with the flake, I didn't see a clean way to pull in the git tag. Maybe a root `VERSION` file? I haven't grokked your build/package workflow yet, but I'll take a look at this when I'm handling the other changes.

ernestrc:
> rune-ide

that works, thanks!

> Oh and how do you want to handle the version string? From my tests with the flake, I didn't see a clean way to pull in the git tag. Maybe a root `VERSION` file? I haven't grokked your build/package workflow yet, but I'll take a look at this when I'm handling the other changes.

A root VERSION file could work, then we'd create release branches to change it from `development` to say `v1.2.2`, and publish the artifacts from the release branch. Leave as it is for now, because this mechanism is the same one I used for other distros, and then we can do follow up one commit to change the way we inject the version in all of them.

PR #103, opened 2026-09-18
Fail compaction when the summary is truncated instead of persisting it

### LINKED ISSUE #78: Compaction persists a length-truncated summary, and the resumed turn degenerates into a closing-tag loop

ejb711:
Fixes #78.

`Summarize` ignored the stream's finish reason, so a summary cut off by the output-token limit was
accepted. `cleanSummary` could not match the unterminated `<summary>` block, and the raw fragment,
dangling tag included, was archived in place of the conversation as the resume prompt. The next
turn was handed an unfinished document and spent itself repairing it, emitting runs of closing tags
instead of resuming the task.

Changes:
- `Summarize` returns an error when the stream finishes with `FinishReasonLength`, naming the
  output-token count and pointing at `/max_tokens`. `CompactDialogue` returns before archiving, so
  the dialogue is kept, the error is surfaced, and the turn continues with its full context. The
  auto-compaction path already logs and continues on a compaction error; the `compact` tool already
  reports the failure and continues without compacting.
- `cleanSummary` tolerates a `<summary>` tag that is never closed and strips a trailing run of
  `</parameter>` / `</invoke>` tags, so a degenerate response cannot leak stray markup into the
  resume prompt.

Not included: the optional retry with a larger budget from the issue. With #80 the session budget
reaches the summary request, which makes truncation rare; this change makes it non-destructive when
it still happens. A retry can follow separately.

Tests (all fail on `main`): `TestCompactDialogueRejectsTruncatedSummary`,
`TestAutoCompact/auto-compact with a truncated summary keeps the full dialogue`, and two
`TestCleanSummary` rows (unclosed tag, trailing tool-call tags).


cmd/rune-agent/agent/agent.go
```diff
@@ -1903,13 +1908,20 @@ func Summarize(
...
+		}
 		if ev.Type == llmapi.EventStreamError {
 			return "", ev.Error
 		}
 	}
 	if err := it.Err(); err != nil {
 		return "", err
 	}
+	if done != nil && done.FinishReason == llmapi.FinishReasonLength {
+		return "", fmt.Errorf("summary truncated after %d output tokens; raise /max_tokens",
```
ernestrc, line 1922: Add a "current max_tokens %d", so the user has a reference point to raise from

cmd/rune-agent/agent/agent.go
```diff
@@ -1903,13 +1908,20 @@ func Summarize(
...
+			done.Usage.TokensReceived)
```
ernestrc, line 1923: it's possible that a provider omits this field and it's just set to 0. I would handle this by printing a different error message maybe without the "after %d" part.

cmd/rune-agent/agent/agent.go
ernestrc, line 1855: "If no <summary> tags are found" is no longer true

PR #82, opened 2026-09-16
Windows graphics Windows graphics: Direct3D 11 + DWM compositor

Bell8571:
Draft first slice of Windows graphics support. Companion issue: https://github.com/unstablebuild/rune/issues/81#issue-5478201845

## This PR

- `internal/term/gui/wingfx` — per-monitor DPI v2, dark title bar, rounded corners, Mica (Win11) / Acrylic (Win10)
- `internal/term/gui/gui_native.go` — `prepareNativeWindow` + `applyNativeWindowEffects` (no-op off Windows)
- `internal/term/gui/font/emoji_windows.go` — Segoe UI Emoji / Noto Color Emoji search paths
- docs + Windows CI skeleton

Wire-up in `gui.go` (two calls only):

```go
// Run, after defer:
prepareNativeWindow(title)

// Update, first line:
applyNativeWindowEffects(g.enableTransparent, g.bgBlurRadius)

Issue #81, opened 2026-09-16
Windows graphics: Direct3D 11 + DWM window effects


Bell8571:
## Why

The README documents a native graphics pipeline as OpenGL on Linux and Metal on macOS. Ebiten already speaks Direct3D 11 on Windows. Rune is missing the platform glue: per-monitor DPI v2, a dark title bar, DWM backdrop (Mica on Windows 11 / Acrylic on Windows 10) so `ScreenTransparent` and blur match the other OSes, and Windows emoji font search paths.

Goal is to land this upstream under GPLv3-or-later so everyone can run Rune as a native Windows app. Not a proprietary fork.

## Approach

Keep the existing Ebiten `Game` loop and character grid. No second renderer.

- `internal/term/gui/wingfx` behind build tags (`windows` / `!windows`)
- Per-monitor DPI v2 before the first HWND
- DWM attributes once the HWND exists (dark mode, rounded corners, Mica / Acrylic when transparency + blur are on)
- `internal/term/gui/font/emoji_windows.go` next to the existing `fonts_windows.go`
- Two calls from the GUI loop: `prepareNativeWindow(title)` in `Run`, `applyNativeWindowEffects(...)` at the start of `Update` (no-ops off Windows)
- Document Direct3D 11 next to OpenGL / Metal

Draft lives on the fork:

- Branch: https://github.com/Bell8571/rune/tree/windows-graphics
- Draft PR on the fork: https://github.com/Bell8571/rune/pull/1

## Known blockers outside this repo

A green `GOOS=windows` build also needs small fixes in the forks Rune already uses:

1. `unstablebuild/ebiten` — `glfw.InitHint` is Unix-only but called unconditionally
2. `unstablebuild/tcell` — `console_win.go` is behind the fork's API

Happy to split those into separate PRs on those repos.

## License

GPLv3-or-later, same file headers as the rest of the tree. Contributors keep copyright.

### Proposed behavior

Windows adoption 

### Area

Install, upgrade, or licensing

### Checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I would be willing to work on this.

ejb711:
We're building a Windows binary from `main` + #82 + the tcell PR 6 and ebiten `InitHint` fixes, cross-compiled from WSL and tested natively on Windows 11. With `SIGUSR1` stubbed, the in-repo tree already compiles for `GOOS=windows`; the two forks are the only blockers.

[...]

Issue #68, opened 2026-09-14
Windows OS Support


vitorfloriano:
### Problem

Okay, I'm gonna be that dev who asks for Windows OS support. 

I understand that Windows s*cks, but I'm kinda forced to use it at work.

Oh, and don't forget WSL support, extremely necessary.

Anyway, is that on the roadmap?

### Proposed behavior

Install Rune on Windows and have it support WSL2.

### Area

Install, upgrade, or licensing

### Additional context

Since [Zed](https://zed.dev/blog/zed-for-windows-is-here) and [Warp](https://www.warp.dev/blog/launching-warp-on-windows) support Windows, perhaps there's something to learn from them.

### Checks

- [x] I searched existing issues and this is not a duplicate.
- [ ] I would be willing to work on this.

ernestrc:
Thanks for creating this issue. It had to be done! We'll prioritize if there's enough 👍 on this issue.

ejb711:
**Windows: file URIs and local paths in rune-go-sdk (design proposal)**

We have `rune.exe` built from `main` + #82 + the open Windows slices, and running natively on Windows 11 up to the bootstrap wizard. Most of what was left after compiling came from one place, file URIs.

**What breaks**
- `workspaceapi.makeLocalURI` builds `"file://" + path`. `D:\work` parses as host `D:` with an invalid port, so Rune exits before handling any flag.
- `uri.go` uses `filepath.Clean/Dir/Base` on URL paths, which turns `/` into `\` (`file://%5CD:...`).
- Callers use `URI.Path()` as an OS path. The file scheme resolved `/C:/Users/me` as its cwd (`file_scheme.go`, 17 sites), and the watcher joined `D:\x\D:\x`.
- Session flags travel inside `syscall.SysProcAttr`, which has no `Setsid`/`Setctty` on Windows. A Windows client can't ask a remote Linux workspace for a controlling tty.

**Proposed SDK API**
- `slashPath`/`localPath` (per OS; identity on unix) map `D:\x` to and from `/D:/x`, the path of `file:///D:/x`.
- `URI.LocalPath()` returns the path as an OS path. `Path()` stays the URL path.
- `FileURI(localPath)` exports the local-path-to-URI conversion so callers stop concatenating `"file://"`.
- URL paths are edited with package `path`, not `filepath`. Nothing changes on unix, where they behave the same.
[...]

PR #106, opened 2026-09-18
Refactor unix-only signal and dup2 handling into build-tagged files

ejb711:
Part of Windows support (#68, coordinated on #81). First of a few small slices that move unix-only
syscalls behind build tags so the tree compiles for `GOOS=windows`; graphics stay with #82.

`cmd/rune/main.go` referenced `syscall.SIGURG` and `unix.Dup2`, and the ssh workspace's signal map
referenced `SIGUSR1`/`SIGUSR2`. None of these exist on Windows, so both packages fail to compile
there.

The unix bodies move into `!windows` files unchanged. Windows gets small counterparts: `SIGURG`
never arrives, the launch log takes over the standard error handle with `SetStdHandle` (which the
runtime also consults for fatal writes, so crash output lands in the log too), and `USR1`/`USR2` are
simply absent from the map.

Behavior on Linux and macOS is unchanged. New tests pin the `SIGURG` check and the forwarded user
signals.

After this, `cmd/rune`'s only remaining Windows compile errors are `detachFromTerminal` and
`presetStandardYAML`, which #82 already provides in `env_windows.go` and
`preset_standard_windows.go`; `workspacessh` has none.


internal/workspace/workspacessh/signal_unix.go
```diff
@@ -0,0 +1,26 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//go:build !windows
+
+package workspacessh
+
+import "syscall"
+
+func init() {
+	sigMap[syscall.SIGUSR1] = "USR1"
```
ernestrc, line 24: rather than mutating the default map, I think it would be better if we just defined the full map here, and a `signal_windows.go` which what we need for windows. Reading std_remote.go no longer tells you what sigMap contains; worse, it leads you to think that that's the entire set.

cmd/rune/signal_windows.go
```diff
@@ -0,0 +1,37 @@
...
+//
+// This program is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//go:build windows
```
ernestrc, line 17: nit: this is redundant. `_windows.go` suffix already implies it

PR #108, opened 2026-09-18
Add procattr to start and stop process trees on each platform

ejb711:
Part of Windows support (#68, coordinated on #81). Third slice; graphics stay with #82.

Four call sites built `syscall.SysProcAttr` literals with `Setpgid`, `Setsid` and `Setctty`, and the
file scheme killed process groups with `syscall.Kill`. None of these exist on Windows, so
`workspace`, `workspacerpc`, `extensionv2`, `idedebug` and `vte` fail to compile there.

This collects them in a small `internal/procattr` package: `NewGroup`, `NewSession`, `LeadsGroup`,
`KillGroup` and `Signal`. The unix implementations are the old code moved unchanged, including
`killProcessGroup` from the file scheme. On Windows a new group is `CREATE_NEW_PROCESS_GROUP`, a
group is killed with `taskkill /T /F` (which walks the process tree, with "not found" mapped to
`os.ErrProcessDone` like `ESRCH` on unix), sessions and controlling terminals do not exist, and
`Signal` supports `SIGKILL` only.

Linux and macOS are unchanged: the existing process-group tests in `file_scheme_test` and
`workspace_runner_test` pass as before, and new tests pin that the helpers build exactly the
attributes the literals built. The Windows-tagged tests cover the `taskkill` arguments and the
group check; `GOOS=windows go vet ./internal/procattr/` passes.

`idelsp`'s single `Setpgid` literal is deliberately left alone here because #107 touches the same
file's imports; it becomes a one-line follow-up once both land.


internal/procattr/procattr_windows.go
```diff
@@ -0,0 +1,71 @@
...
+// LeadsGroup reports whether attr starts the process as the root of a new
+// process group.
+func LeadsGroup(attr *syscall.SysProcAttr) bool {
+	return attr != nil && attr.CreationFlags&syscall.CREATE_NEW_PROCESS_GROUP != 0
+}
+
+// KillGroup terminates proc and every process descended from it.
+func KillGroup(proc *os.Process) error {
+	// Windows has no group-wide signal; taskkill /T walks the parent-pid tree.
+	err := exec.Command("taskkill", taskkillArgs(proc.Pid)...).Run()
```
ernestrc, line 52: Consider adding `SysProcAttr{HideWindow: true}` so every kill doesn't flash a console window.

internal/procattr/procattr_windows.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// LeadsGroup reports whether attr starts the process as the root of a new
+// process group.
+func LeadsGroup(attr *syscall.SysProcAttr) bool {
+	return attr != nil && attr.CreationFlags&syscall.CREATE_NEW_PROCESS_GROUP != 0
+}
+
+// KillGroup terminates proc and every process descended from it.
+func KillGroup(proc *os.Process) error {
```
ernestrc, line 50: I would also thread a context here and pass it as `exec.CommandContext`. On unix it's just a syscall, but on windows we're spawning a subprocess that can block for an unbound amount of time.

internal/procattr/procattr_windows.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func taskkillArgs(pid int) []string {
+	return []string{"/T", "/F", "/PID", strconv.Itoa(pid)}
+}
+
+// Signal sends sig to the process with the given pid. Windows supports only
+// SIGKILL.
+func Signal(pid int, sig syscall.Signal) error {
+	proc, err := os.FindProcess(pid)
```
ernestrc, line 66: I think we need a `defer proc.Release()` here.

internal/procattr/procattr_unix_test.go
```diff
@@ -0,0 +1,75 @@
...
+		})
+	}
+}
+
+func TestKillGroupAfterExitReportsProcessDone(t *testing.T) {
+	cmd := exec.Command("true")
+	cmd.SysProcAttr = NewGroup()
+	require.NoError(t, cmd.Run())
+
+	assert.ErrorIs(t, KillGroup(cmd.Process), os.ErrProcessDone)
```
ernestrc, line 74: After cmd.Run() the process has returned, the pgid is free for reuse and SIGKILL(-pid) can land on an unrelated group. Low probability, but getting random process killed on a busy CI machine would not be fun to debug.

internal/workspace/file_scheme.go
```diff
@@ -429,10 +430,9 @@ func (p *fileScheme) StartCommand(ctx context.Context, cmd workspaceapi.Cmd) (
 	// `go run` exec the real program as a grandchild that SIGKILL
 	// cannot be forwarded to. Joining an existing Pgid is someone
 	// else's group and not ours to signal.
-	attr := cmd.SysProcAttr
```
ernestrc, line 432: the comment above now belongs to the LeadsGroup implementations

internal/workspace/file_scheme.go
```diff
@@ -518,22 +518,13 @@ func (p *fileScheme) Signal(pid workspaceapi.Pid, signal syscall.Signal) error {
 		return errProcNotFound
 	}
 
-	err := syscall.Kill(int(pid), signal)
+	err := procattr.Signal(int(pid), signal)
 	if err != nil {
 		return fmt.Errorf("syscall kill: %w", err)
```
ernestrc, line 523: we should update this error now to "Send signal: %w"

PR #112, opened 2026-09-18
Fix firstmover lock takeover on Windows

ejb711:
Part of Windows support (#68, coordinated on #81). Found by running a locally built `rune.exe` on
Windows 11; graphics stay with #82.

The firstmover lock is a unix socket. When a leader dies without removing it, the next process gets
an address-in-use error from `bind(2)` and falls back to following, then to a coup once the dead
leader stays unresponsive. That check used `syscall.EADDRINUSE`, which on Windows is an invented Go
errno rather than the Winsock `WSAEADDRINUSE` that bind returns, so the error was treated as
unexpected and the service gave up. After any run that did not exit cleanly, every storage user
failed on the next launch: session history, window size and position, installed packages.

Windows now also matches `WSAEADDRINUSE`, and the lock directory is created before the first bind,
since Windows reports a missing parent directory as `WSAENETDOWN` rather than `ENOENT` (same shape
as #111).

Linux and macOS are unchanged. New tests cover the per-OS check and, end to end, both a stale socket
left by a dead leader and a lock directory that does not exist yet. Verified natively: relaunching
on a data directory holding stale locks now logs the normal "Unresponsive leader, starting coup"
recovery instead of seven storage errors.


internal/localstorage/firstmover/service.go
```diff
@@ -852,18 +857,9 @@ func (s *Service) leadOrFollow() {
 			return true, err
 		}
 
-		if errors.Is(err, syscall.ENOENT) { // a component of the path does not exist
```
ernestrc, line 855: Moving MkdirAll out of the retry loop means the lock directory is created exactly once, at goroutine start. ENOENT is now absent from the expected-error set at line 862-864, so if Listen ever returns it, control falls into:
```go
s.log(log.WarnLevel, "Unexpected error while trying to acquire lock %q: %v", ...)
return false, err
```

retry.Retry stops permanently on (false, err) (internal/extension/extensionv2/runner.go:244 doc), readyErr is set, and every subsequent Get/Set fails for the process lifetime. Before this PR the same
condition recreated the directory and retried.

Let's keep the MkdirAll but put it back inside fn, before cfg.Listen. It restores self-healing on
both platforms, and it still satisfies the Windows constraint the comment describes

PR #114, opened 2026-09-18
Fix terminal modal mode for bash and fish via bell bindings

### LINKED ISSUE #98: Fix terminal modal (vi) mode for bash and fish via bell bindings
## Summary

Modal (vi) mode in Rune terminals only works with zsh. Under **bash** (including `set -o vi`) and **fish** it never engages: the vte synchronizes with the shell by writing `^A^G` to the pty and waiting for the echoed BEL (`internal/term/vte/remote.go:154`, `wait_parser_handler.go:109`). Bash's vi-insert keymap `self-insert`s both bytes (hence the visible `^A^G`), and fish binds neither.

Both fixes below were verified empirically through a pty (bash 3.2 emacs/vi, runtime `set -o vi`, `.inputrc` `editing-mode vi`; fish 4.2 default, vi mode, user rebinds in `config.fish`).

## Design decisions

1. **bash → `INPUTRC` injection.** Ship an `inputrc` next to the zsh dotfiles in the existing bundled `zdot/` directory and export `INPUTRC=<zdotdir>/inputrc` when the resolved login shell is bash. This reuses all existing plumbing (bundle → `installZdotDir` → `--rune-zdotdir` → config `zdotdir` → `fileScheme.zdotDir`, including the remote-config path for SSH), so no new flag/option/config key. Guard with `os.Stat`: readline does **not** fall back to `~/.inputrc` when `$INPUTRC` points at a missing file, so a stale/custom `zdotdir` must not silently eat the user's bindings.
   - Alternative (not chosen): a separate `inputrc` config key + `--rune-inputrc` flag + `ide.WithInputrc`. More explicit but ~6 parallel plumbing touches for no functional gain.
2. **fish → `-C`/`--init-command`.** No file to ship. Runs after `config.fish`, so it beats user rebinds and survives `fish_vi_key_bindings` (which only erases `--preset` bindings). Applied only when `StartCommand` defaults the args (interactive login-shell contract), matching where `--login -i` is chosen. Uses legacy `\ca`/`\cg` notation, valid on fish 3.x and 4.x.

### 1. New bundled file `extra/osx/Rune.app/Contents/Resources/zdot/inputrc`

```
# Rune terminal readline bindings; see internal/workspace/file_scheme.go.
$include /etc/inputrc
$include ~/.inputrc
$if Bash
set bell-style audible
set keymap vi-insert
"\C-a": beginning-of-line
"\C-g": abort
set keymap vi-command
"\C-g": abort
set keymap emacs
"\C-a": beginning-of-line
"\C-g": abort
$endif
```

- Binding each keymap unconditionally (not `$if mode=vi`) is what makes runtime `set -o vi` work: keymaps are global and readline restores the active one from the edit mode after parsing.
- `$if Bash` keeps the override away from other readline programs (python, gdb, psql) that inherit `INPUTRC` from the Rune shell.
- macOS bundling is automatic (`cp -fRp $(APP_TEMPLATE)`).

### 2. `cmd/rune/main.go`

- `zdotFiles`: append `"inputrc"` so `installZdotDir` mirrors it into `<datadir>/zdot`. Existing `TestInstallZdotDir*` tests iterate `zdotFiles` and cover it.

### 3. Linux packaging

- `cmd/rune/Makefile` `make-release-linux`: add `cp .../zdot/inputrc .../rune.app/share/zdot/`; update the layout comment at line 364.
- `deploy/rune-linux/Dockerfile` (~line 206): same `cp`.
- `deploy/rune-linux/README.md` layout block: list `inputrc` under `zdot/`.

### 4. `internal/workspace/file_scheme.go` — `StartCommand` (lines 383-391)

Replace the zsh-only branch with a switch on `filepath.Base(cmd.Path)` inside the empty-Path block:

```go
argsDefaulted := len(cmd.Args) == 0
if argsDefaulted { cmd.Args = []string{"--login", "-i"} }
switch filepath.Base(cmd.Path) {
case "zsh":
    if p.zdotDir != "" { cmd.Env = append(cmd.Env, "ZDOTDIR="+p.zdotDir) }
case "bash":
    if p.zdotDir != "" {
        inputrc := filepath.Join(p.zdotDir, "inputrc")
        if _, err := p.osStat(inputrc); err == nil {
            cmd.Env = append(cmd.Env, "INPUTRC="+inputrc)
        }
    }
case "fish":
    if argsDefaulted { cmd.Args = append(cmd.Args, "-C", FishInitCommand) }
}
```

Add:

```go
// FishInitCommand binds ^A/^G in fish so the vte's bell handshake
// (see vte.ptyWriter.triggerBell) works: fish binds neither by default
// and has no beep widget, so ^G prints BEL itself.
const FishInitCommand = `bind \ca beginning-of-line; bind \cg 'printf \a'; ` +
    `bind -M insert \ca beginning-of-line; bind -M insert \cg 'printf \a'`
```

Exported so the vte e2e test can reuse it. Use `p.osStat` (already a field) so tests can stub it. Update the `zdotDir` field comment to say it also hosts the bash `inputrc`.

### 5. Tests

**Hermetic — `internal/workspace/file_scheme_test.go`** (same trampoline pattern as the existing zsh subtests, `SHELL` pointed at a script whose basename is `bash`/`fish`):

- `bash exports INPUTRC when <zdotdir>/inputrc exists` — create `zdotdir` tmp with an `inputrc`; `-c 'echo INPUTRC=$INPUTRC'` → `INPUTRC=<zdotdir>/inputrc`.
- `bash leaves INPUTRC untouched when inputrc is missing` — same without the file; `t.Setenv("INPUTRC","")`; expect `unset`.
- `fish defaults args to --login -i -C <FishInitCommand>` — trampoline `printf '%s\n' "$@"`; `Args` nil; assert stdout lines equal `--login`, `-i`, `-C`, `FishInitCommand`. (The trampoline only echoes args; it doesn't need to be fish.)
- `fish with explicit Args gets no -C`.
- Existing "without zsh leaves ZDOTDIR untouched" test keeps passing.

**e2e — `internal/term/vte`** (`//go:build e2e`, skip when the shell isn't installed, like `TestZshEdgeCases`):

- `handler_test.go`: generalize `testSequenceShell` into `testSequenceCommand(t, cfg, timeout, commandAndArgs []string, cases)`; keep `testSequenceShell` as a one-line wrapper.
- `vi_test.go` `TestBashViModeEdgeCases`: reproduces the bug (TDD). Temp rcfile with `set -o vi` and `PS1='$ '`; `os.Setenv("INPUTRC", "<repo>/extra/osx/Rune.app/Contents/Resources/zdot/inputrc")` (path relative to the package dir); command `[bash --noprofile --rcfile <rc> -i]`; `cfg.Modal = true`; a case like `"echo blaaa<0Cecho hi"` expecting the prompt row `$ echo hi▐`. Without the inputrc the sequence shows `^A^G` and never enters modal mode, so the test fails before the fix.
- `vi_test.go` `TestFishEdgeCases`: command `[fish --no-config -i -C "<workspace.FishInitCommand>; function fish_prompt; printf '$ '; end"]`; same style case. Risk to check while running it: fish 4 probes DA1/CPR at startup (`\x1b[0c`, `\x1b[6n`); if the vte doesn't answer, fish proceeds after its own short timeout, so the first case may need a slightly larger idle wait.

### 6. Out of scope / notes

- `ideConfig.zdotDir` (`internal/ide/config.go:265`) is write-only dead plumbing; left alone.
- Users who disable the bell (`bind 'set bell-style none'` in `.bashrc`, `unsetopt beep`) still hit the existing "modal mode cannot be enabled" warning; unchanged.

## Verification

```
go test -race ./internal/workspace/ ./cmd/rune/
go test -race -tags e2e -run 'TestBashViModeEdgeCases|TestFishEdgeCases|TestZshEdgeCases' ./internal/term/vte/
make lint && make test
```

Manual: launch Rune with `SHELL=/bin/bash`, run `set -o vi`, press `<esc>` → modal mode engages, no `^A^G` echo; same with `SHELL=$(which fish)`.

ejb711:
Fixes #98. The design and the `inputrc` are taken from the issue as written.

Modal mode only engaged under zsh. The vte synchronizes with the shell by writing `^A^G` to the pty
and waiting for the bell, but bash's vi-insert keymap self-inserts both bytes, and fish binds
neither, so the handshake never completed and `^A^G` showed up in the command line.

- **bash:** an `inputrc` ships next to the bundled zsh dotfiles, and `INPUTRC` is exported when the
  login shell is bash and the file exists (readline does not fall back to `~/.inputrc` when
  `INPUTRC` names a missing file). It includes the system and user files first and binds `^A`/`^G`
  in every keymap under `$if Bash`, so runtime `set -o vi` works and other readline programs are
  unaffected. Installed with the other zdot files and added to the Linux bundle.
- **fish:** the bindings are passed with `--init-command` when the shell starts with the default
  login arguments; they run after `config.fish` and survive `fish_vi_key_bindings`.

Tests: hermetic cases in `TestStartCommand` for `INPUTRC` (present / missing) and the fish
arguments (defaulted / explicit); e2e cases (`-tags e2e`) drive bash and fish through a real pty in
modal mode, alongside the existing zsh case as a control. Verified against bash 5.1.16, fish 4.9.3
and zsh 5.8.1. Before the change the bash e2e case shows the literal `^A^G` in the prompt and the
fish case never enters modal mode.


zhengqunkoo:
Can confirm this fix works for bash

internal/workspace/file_scheme.go
```diff
@@ -382,11 +389,28 @@ func (p *fileScheme) StartCommand(ctx context.Context, cmd workspaceapi.Cmd) (
...
+			if p.zdotDir != "" {
+				cmd.Env = append(cmd.Env, fmt.Sprintf("ZDOTDIR=%s", p.zdotDir))
+			}
+		case "bash":
+			// readline does not fall back to ~/.inputrc when INPUTRC names a
+			// missing file, so only export one that exists.
+			if p.zdotDir != "" {
+				inputrc := filepath.Join(p.zdotDir, "inputrc")
+				if _, err := p.osStat(inputrc); err == nil {
+					cmd.Env = append(cmd.Env, "INPUTRC="+inputrc)
```
ernestrc, line 407: a bash terminal exports INPUTRC to all children, but a zsh terminal exports only ZDOTDIR, so running bash inside a Rune zsh terminal gets no bindings and modal mode breaks there.

Not sure what the best solution is here, maybe setting both env vars in both cases?

internal/workspace/file_scheme.go
```diff
@@ -382,11 +389,28 @@ func (p *fileScheme) StartCommand(ctx context.Context, cmd workspaceapi.Cmd) (
...
+		switch filepath.Base(cmd.Path) {
+		case "zsh":
+			if p.zdotDir != "" {
+				cmd.Env = append(cmd.Env, fmt.Sprintf("ZDOTDIR=%s", p.zdotDir))
+			}
+		case "bash":
+			// readline does not fall back to ~/.inputrc when INPUTRC names a
+			// missing file, so only export one that exists.
+			if p.zdotDir != "" {
+				inputrc := filepath.Join(p.zdotDir, "inputrc")
```
ernestrc, line 405: If anyone sets the (undocumented) `workspace.file.zdotdir` in their `.rune/config.yaml` to their own zsh config dir, then `p.zdotDir` is set, but the inputrc will probably not be there, because it's only for bash, so users won't have an inputrc there.

I understand the motivation behind piggybacking off zdotdir, but I think we can cut a line and only use it to bundle and pass the inputrc file, and then have another field in the scheme for the location of inputrc.

internal/workspace/file_scheme.go
zhengqunkoo: My two cents:

Even though nested shells seems like an edge usage pattern, I don't see an issue with exporting both env vars.

I would add these tests:

1. nested bash receives `INPUTRC`;
2. nested zsh receives `ZDOTDIR`;

PR #115, opened 2026-09-18
Fix PATH setup on Windows and skip the login-shell PATH probe there

ejb711:
Part of Windows support (#68, coordinated on #81). Found by running a locally built `rune.exe` on
Windows 11; graphics stay with #82.

`setRuneBinPATH` joined the managed bin directory onto `PATH` with `:`, which on Windows fuses it
with the first entry instead of adding a new one. It now joins with `os.PathListSeparator` through
a small pure helper; on Linux and macOS the result is byte-identical to before.

The login-shell `PATH` probe runs a POSIX script with `-i -l -c` and `/usr/bin/env`, which cannot
succeed on Windows and spent up to its 10 s timeout on every launch. It moves to a `!windows` file
unchanged; on Windows, which has no login shell that extends `PATH`, the current `PATH` is used.

Tests: `TestPrependPATH` covers both separators and the empty base; the existing PATH test no longer
assumes `:`; a Windows-tagged test pins that the shell probe is never consulted there. The files are
named `login_path_*` to avoid an add/add conflict with #82's `env_windows.go`; once this lands,
#82's `detachFromTerminal` stub is no longer needed.


cmd/rune/env_test.go
```diff
@@ -269,6 +269,24 @@ func TestResolveLoginPathFallsBackShell(t *testing.T) {
 
 // TestSetupManagedBinPathPrependsBinDirOnce asserts that setupManagedBinPath
 // creates the managed dirs and prepends the bin dir to PATH exactly once.
+func TestPrependPATH(t *testing.T) {
```
ernestrc, line 272: the comment above is incorrect now. Move this test 3 lines up

PR #116, opened 2026-09-18
Fix key-binding migration corrupting flow-style configs

ejb711:
Found while testing a config written as `{}`.

The key-binding migration splices the preset's bindings into a pre-preset config line by line,
which assumes block YAML. A config written as a flow mapping, even an empty `{}`, ended up with a
block-style command section appended after the flow mapping and a version stamp in front of it:
an invalid document, so Rune started without the user's config
(`yaml: line 3: could not find expected ':'`).

A flow-style top level is now rewritten as block YAML before splicing; an empty `{}` becomes an
empty document. Block-style configs, the common case, pass through untouched so their formatting
and comments survive exactly as before (the existing idempotency and "leaves shipped presets
alone" tests cover that).

Test: `TestMigrateConfigKeyBindingsHandlesFlowStyleConfig`, a table with `{}`, a flow config with
settings that must survive, and one selecting the emacs preset. All three fail on `main` with the
error above. Known limit: a trailing comment on the flow line itself is dropped in the conversion.

This merges cleanly with #90, which changes `renderPreset` to take a second argument; if #90 lands
first, the new test's `renderPreset(tc.editor)` call needs `, true` like the others.


ernestrc:
I forgot this even existed. This was a migration that had to happen a while ago, when rune.star carried defaults and the presets had just "overrides". It's not needed anymore.

ernestrc:
> I forgot this even existed. This was a migration that had to happen a while ago, when rune.star carried defaults and the presets had just "overrides". It's not needed anymore.

Removing it here: https://github.com/unstablebuild/rune/commit/fd9d566dac56d53e3d96b554e6b85e9f069cc149

PR #117, opened 2026-09-18
Fix GUI freeze on terminal suspend signals

### LINKED ISSUE #100: Rune GUI window freezes and becomes uncloseable when terminal process is stopped/suspended
### What happened

When launching Rune from a terminal (`./bin/rune`), stopping or suspending the process from the terminal (e.g. with `Ctrl+Z` or closing/disconnecting the terminal) causes the Rune GUI window to freeze permanently on the desktop. 

The frozen window cannot be interacted with, does not respond to mouse clicks, and ignores window manager shortcuts (such as `Super+W` or `Alt+F4`). Running `ps aux` reveals the Rune process enters state `Tl` (stopped by job control), completely freezing the GUI event loop. The only way to close the window is to manually locate the PID and run `kill -9 <PID>`.

Additionally, when Rune is terminated this way, spawned extension processes (`rune-agent`, `extension_python`, `extension_go`, etc.) remain orphaned in the background.

### What you expected

1. Suspending or terminating the terminal process should not leave an unresponsive zombie GUI window on the desktop. Either terminal signals (SIGTSTP/SIGINT/SIGHUP) should be handled gracefully or cleanly shut down the GUI and terminate child extension processes.
2. A clear, documented global shortcut to cleanly quit Rune (e.g., `Ctrl+Q` or `Super+Q`) without requiring the command palette (`:quit`) or hunting for PIDs.

### Steps to reproduce

1. Launch Rune from a terminal: `./bin/rune`
2. Switch focus back to the terminal and press `Ctrl+Z` (or attempt to interrupt/stop the process).
3. Switch back to the Rune window (or switch to its virtual workspace).
4. Attempt to close or interact with the window (clicking the close button, pressing `Super+W`, `Esc`, or typing).

**Observed:**
The window is completely frozen and cannot be closed or focused. Checking `ps aux | grep rune` shows `./bin/rune` in `Tl` state along with running child extension processes.

### Area

GUI and rendering

### Rune version

Rune v1.2.1-28-gc1e83721

### OS and architecture

Linux x86_64 (Wayland / X11)

### Shell

bash 5.3

### Logs, crash report, or screenshots

```shell
$ ps aux | grep -i rune
spandev  651062 42.9  6.4 8761484 1044360 pts/3 Tl   17:29  10:26 ./bin/rune
spandev  651425  0.0  0.2 2040848   33572 pts/3 Sl   17:29   0:00 /home/spandev/.rune/bin/rune-agent
spandev  651434  0.0  0.1 2020868   28248 pts/3 Sl   17:29   0:00 /home/spandev/.rune/bin/extension_python
spandev  651441  0.0  0.1 2172560   31136 pts/3 Sl   17:29   0:00 /home/spandev/.rune/bin/extension_go
```

### Relevant configuration

```yaml

```

### Checks

- [x] I can reproduce this on the latest released version.

JoeJoeflyn:
### Summary

GUI mode installed no signal handler, so stopping the process from the launcher terminal (Ctrl+Z, or closing the terminal) left a permanently frozen, uncloseable window on the desktop, and killed runs orphaned extension child processes. This adds a small GUI-mode signal watcher:

- `SIGINT`, `SIGTERM`, `SIGHUP` publish the same quit event used by window close, so the app exits through the normal handler chain and deferred closes terminate extension processes
- `SIGTSTP` is ignored: a stopped GUI client cannot answer the compositor's close requests, which is what produced the zombie window

### Implementation Details

- `watchGUISignals` (new `signals_unix.go`, `!windows` tag) registers the handlers and returns a cleanup func; wired as a deferred call in `runGUI` after `attachGUI` so `publishEvent` can deliver
- Terminate signals reuse the existing `quitEvent(appMenuKeyBindings(cfg))` path — identical to the window ✕ / Quit menu — so `ErrHandlerExited` runs `g.Close()` / `root.Close()` / `runner.Close()` and extensions die cleanly instead of orphaning
- Windows stub is a no-op: no POSIX job-control signals exist there
- Tests drive the real path: `syscall.Kill(getpid(), SIGTERM)` asserts the quit event is published; `SIGTSTP` asserts the signal is ignored (checking `SigIgn` in `/proc/self/status` on Linux) instead of stopping the test process

### Verification

- `go test -race ./cmd/rune/` — green, including both new tests
- Manual repro of the issue steps on Linux/Wayland: launched `./bin/rune -G` from a shell in a real pty — before this change `Ctrl+Z` put the process in `Tl` state with a frozen uncloseable window; after, the process stays `Sl+` and responsive, and `Super+W` / ✕ close it normally
- `kill -TERM` on the running GUI exits through the quit path (default disposition is bypassed once `signal.Notify` is registered)

One follow-up noted but out of scope: children orphaned by uncatchable kills (`SIGKILL`, stop→kill) would need a stdin-EOF watchdog in `rune-go-sdk`'s serve loop — happy to file that separately.

Fixes #100

screenrecording-2026-09-19_00-30-12


Spandan-Mishra:
i tested this out on my system too and `Ctrl+Z` & `Ctrl+C` in the terminal work as expected :+1:

cmd/rune/signals_unix.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"unstable.build/rune/internal/debug"
+)
+
+// watchGUISignals publishes the quit event on terminal signals so the
+// GUI exits through the same close chain as a window ✕. SIGTSTP is
+// ignored instead: a stopped client cannot answer the compositor's
+// close requests, leaving a frozen window.
+func watchGUISignals(publish func(term.Event) bool, quit term.Event) func() {
+	signal.Ignore(syscall.SIGTSTP)
```
ernestrc: this also forces child processes to ignore SIGSTP. Instead of doing it here, why not listen for it below, then ignore it just for this process:

```go
  if sig == syscall.SIGTSTP {
        continue
    }
```

cmd/rune/signals_unix.go
```diff
@@ -0,0 +1,58 @@
...
+				publish(quit)
+			case <-done:
+				return
+			}
+		}
+	})
+	return func() {
+		close(done)
+		signal.Stop(ch)
+		signal.Reset(syscall.SIGTSTP)
```
ernestrc: then you can remove this as well

cmd/rune/signals_unix.go
```diff
@@ -0,0 +1,58 @@
...
+	signal.Ignore(syscall.SIGTSTP)
+	ch := make(chan os.Signal, 1)
+	signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP)
+	done := make(chan struct{})
+	go debug.CapturePanicReport(func() {
+		for {
+			select {
+			case _, ok := <-ch:
+				if !ok {
+					return
```
ernestrc, line 46: `publish(quit)` calls `GUI.PublishEvent`, which is a non-blocking send that returns false when `updateChan` is full (unlikely but not impossible). Then the result is discarded, so a signal i.e.`SIGTERM` arriving while the event queue is saturated is silently lost and the process never exits.

Consider tracking that a terminate signal was already seen, and on a second signal (or if publish returns false) escalate to an immediate exit.

ernestrc (approved):

PR #119, opened 2026-09-20
Fix clickable markdown links in agent dialogue TUI

### LINKED ISSUE #7: Rune Agent: make markdown links clickable in the dialogue TUI
Markdown links rendered in the Rune Agent dialogue TUI are not clickable. Clicking one starts a text selection instead of following the link.

The SDK already supports link clicks: the markdown handler accepts an `OnLinkClick` option and the markdown component can hit-test a coordinate to a link. The dialogue TUI simply never wires clicks to its markdown components.

## Current state

- All mouse events route through the dialogue's mouse delegate, whose `OnAction` always returns `false`. It only handles selection and scrolling on a flattened cell grid.
- Sent messages and prompt bodies render a bare markdown component with no handler at all.
- Streamed received messages do wrap markdown in the SDK markdown handler, but without a link-click option, and the delegate never invokes that handler's `OnAction`.

## Proposal

1. Accept a link-click callback in the component config (`func(*url.URL) bool`), documented so that nil ignores clicks and the package stays service-agnostic.
2. Add a hit-test that maps a click coordinate to the specific markdown component in the message list and asks it for the link at that point.
3. Handle left clicks in the mouse delegate: resolve a link, parse it, and invoke the callback. Returning `true` suppresses selection-start so the click does not also begin a drag.
4. Wire the real callback at the extension's dialogue construction site, copying http/https URLs to the clipboard and notifying, mirroring the equivalent behavior in the IDE.

Local `#anchor` scrolling is out of scope; bare fragment links should fall through harmlessly.

## Where to look

- `cmd/rune-agent/dialogue/dialoguetui/config.go` - where the new config field belongs, next to the existing markdown config.
- `cmd/rune-agent/dialogue/dialoguetui/component.go` - the three markdown render sites (sent message, streamed receive, prompt body) and the message-list iteration patterns to copy for the hit-test.
- `cmd/rune-agent/dialogue/dialoguetui/mouse_delegate.go` - the delegate and its `OnAction`.
- `cmd/rune-agent/dialogue/dialoguetui/handler.go` - delegate construction, and the place where the message-area offset is already subtracted from the event, so the delegate's coordinates are already in message-viewport space.

## Notes for contributors

Tests belong in the existing mouse test files, using a capturing callback. Worth covering: a click on a link fires the callback and returns true; a click on plain text in the same message does not and returns false so selection still starts; links work in all three render sites; out-of-bounds clicks return false.

Spandan-Mishra:
### Summary
Closes #7.

Fixes an issue in the `rune-agent` dialogue TUI where left-clicking markdown links in sent messages, received messages, or prompt bodies fell through to normal text selection rather than activating the link.

### Changes
- **`cmd/rune-agent/dialogue/dialoguetui`**:
  - Added `OnLinkClick func(*url.URL) bool` to `ComponentConfig`.
  - Implemented `(*Component).LinkAt(pos term.Coordinates) *markdown.LinkInfo` to traverse visible messages in `c.messages`, unwrap spans/handlers, and hit-test underlying markdown components.
  - Updated `mouseDelegate.OnAction` to query `d.comp.LinkAt(pos)` on `mouse.LeftClick`, parse the URL, and invoke `OnLinkClick`, returning `true` to suppress selection.
- **`internal/handler/markdown`**:
  - Exported `Component() *markdown.Component` on `*Handler` to permit component unwrapping during layout hit-testing.
- **`cmd/rune-agent/extension`**:
  - Wired `ret.cfg.OnLinkClick` via `newLinkClickHandler`: copies `http` and `https` URLs to `clipboard.Register` and dispatches a success notification via `browserapi.Notifications`.
- **Tests**:
  - Added regression test `TestMouseDelegate_LinkClick` covering sent messages, received messages, prompt bodies, non-link text fallthrough, unhandled bare anchors, and nil callbacks.
  - Added `TestExtensionDialogue_LinkClick_CopiesToClipboard` testing URL copying, notifications, and non-http filtering.

### Results

https://github.com/user-attachments/assets/942ece2f-a93e-47b8-a1c7-60fa8cf02318


cmd/rune-agent/dialogue/dialoguetui/config.go
```diff
@@ -52,7 +53,7 @@ type ComponentConfig struct {
 	// messages and control their alignment.
 	QueuedMessageSpanConfig component.SpanConfig
 	// QueuedMessagePrefix is the prefix shown before each queued message.
-	// Defaults to "󰄝 " when empty.
+	// Defaults to "󰑝 " when empty.
```
ernestrc: what's this?

PR #122, opened 2026-09-20
Fix resource cleanup paths and add GoHawk to CI

kojah:
Changes:

- Release the token cache mutex before returning from a canceled storage lookup.
- Close telemetry responses and crash report files on every return path.
- Close tar and gzip writers when walking a worktree fails.
- Parse font files from memory instead of retaining open file descriptors.
- Add the pinned GoHawk v0.3.4 binary to the existing lint job, with existing noisy categories disabled so the initial rollout passes.

Testing:

- `go test ./auth ./cmd/rune/ide/apiclient ./internal/debug ./internal/ide/gitpkg`
- `xvfb-run -a go test ./internal/term/gui/font`
- `gohawk -json -enable-all -disable=exitpolicy,goroutineownership,lockorder,processownership ./...`

[gohawk](https://gohawk.dev) found these bugs while I was scanning the repo. I wrote it, so I took the liberty of adding it to the existing CI job along with the fixes. You're very welcome to remove that part if you'd rather not add another tool :)


ernestrc:
@kojah totally the opposite, I want this in our pre-commit hooks. For this to get merged though, we need to have it pass; currently I'm getting many errors. For this PR add all of the `-disable=<check>` such that it passes, and then in subsequent PRs we can incrementally fix categories of errors, thus removing a `-disable<check` per PR. What do you think?

[gohawk.json](https://github.com/user-attachments/files/32443642/gohawk.json)

kojah:
@ernestrc Awesome! I'll work on fixing up the PR so there's less analyzer noise, then, and submit those other ones incrementally like you said. I think the goroutineownership (and perhaps others) can be turned off.

kojah:
@ernestrc Addressed your feedback! I also fixed an issue with a leaked file in the font loader, since it's one of the tinier changes. Let me know if there's anything else I need to do before merge, otherwise I'll wait and submit the other fixes

ernestrc (approved):

PR #123, opened 2026-09-21
Update swap-based flushing to use centralized editor.swap_dir directory

### LINKED ISSUE #91: Add a configurable swap directory, defaulting to a central ~/.rune/swap
## Summary

Rune writes its swap file next to the file being edited: editing `src/main.go`
creates `src/.main.go.rswp`. Every user and every repository Rune touches has to
add `*.rswp` to `.gitignore`, the sibling entries pollute file listings and
search results, and opening a file in a directory the user cannot write fails
outright because the swap cannot be created.

Make the swap location configurable with two supported layouts, and default to
the new one:

| `editor.directory` | Layout | Entry for `/Users/x/src/proj/main.go` |
| --- | --- | --- |
| `"~/.rune/swap//"` (default) | central, per host | `~/.rune/swap/%Users%x%src%proj%main.go.rswp` |
| `"."` | alongside the file (today's behavior) | `/Users/x/src/proj/.main.go.rswp` |

The sibling layout is **not** removed — it stays a first-class, supported option.

## Configuration

A single string key under `editor`, the config object that backs the `text`
component (`editor.tabspaces`, `editor.auto_save`, ... — `cmd/rune/rune.star:347`):

```starlark
"editor": {
    # Where swap files are kept.
    #
    #   "."               keep the swap next to the edited file, as
    #                     ".<name>.rswp". Visible to git; needs write
    #                     permission on the file's directory.
    #   "<path>//"        keep swaps in one directory. The trailing "//"
    #                     marks the mangled-name form: the file's full path
    #                     is encoded into the entry name with "%" separators,
    #                     so files of the same name never collide.
    #
    # The path is resolved on the host that owns the file, so remote
    # workspaces keep their swaps on the remote machine.
    "directory": "~/.rune/swap//",
},
```

Naming follows the convention users will recognize from other editors: a
`directory` key, `"."` for "next to the file", and a trailing `//` marking
full-path-mangled entry names. A central path written without the `//` suffix is
accepted and treated as if it had one — basename-only entries in a shared
directory would collide with no upside.

## Current behavior

A swap directory is already plumbed end to end — only the *policy* function
places the swap next to the file.

`internal/workspace/file.go:142`

```go
func swapFileName(swapDir, filePath string) (string, string) {
	if swapDir == "" {
		swapDir = filepath.Dir(filePath)
	}
	return swapDir, path.Join(swapDir, "."+filepath.Base(filePath)+SwapFileExtensionName)
}
```

Flow: `text.Component.getSwapDir` (`internal/text/component.go:127`) →
`workspace.DefaultSwapDirectory` (`internal/workspace/uri.go:59`, returns
`filepath.Dir(file)`) → `Workspace.Load(file, buf, swapDir, readOnly)`
(`internal/workspace/workspace.go:55`) → `schemeWorkspace.Load`
(`internal/workspace/scheme_workspace.go:125`) → `newFile(..., swapDirPath, ...)`.

`getSwapDir` has four call sites: `internal/text/component.go:175` and
`internal/text/prompts.go:52,113,128` (the recovery prompts, which recompute the
swap path through `workspace.DefaultSwapFile`).

### 1. Save is implemented as a rename of the swap over the original

`internal/workspace/file.go:935`

```go
err = f.scheme.Rename(f.swapFileName, origTarget)
```

`fileScheme.Rename` is `os.Rename` (`internal/workspace/file_scheme.go:303`),
which fails with `EXDEV` across filesystems. Under the central layout **every
save of a file on a different volume would break**: external drives, `/tmp`, a
separate `/home` partition, bind mounts, network volumes, containers.

The rename also donates the swap's inode to the original, so mode, ownership,
ACLs, xattrs and hard links are replaced by the swap's. That is mostly benign
while the swap is a sibling, and wrong once it lives on another filesystem.

**Resolution: keep the rename as a fast path and fall back to an in-place
write.** The rename is an optimization — it avoids writing the bytes twice — not
a correctness requirement: `copyFlushSwapFile`
(`internal/workspace/file.go:473-516`) already truncates, writes the complete
buffer and `Sync`s the swap on every edit, so the swap is self-contained at all
times and a partially written original is always recoverable from it.

So: attempt `Rename`; on any failure write `f.content` into the original in place
(truncate + write + `Sync`) and keep the swap as the journal. There is no second
staged file to maintain. Under `directory = "."` the rename keeps succeeding, so
that layout behaves exactly as it does today. Where the fallback does engage it
is strictly better behaved than the rename it replaces:

- the original keeps its inode, so mode, ownership, ACLs, xattrs and hard links
  survive the save;
- the swap is not consumed by the save, so the `O_EXCL` lock is held
  continuously. Today the rename destroys the swap and `initFiles`
  (`internal/workspace/file.go:944`) recreates it, leaving a window in which the
  file holds no lock at all.

The ordering guarantee must be preserved on both paths: the swap has to be
durable before the original is touched.

### 2. A flat directory needs path mangling

`.main.go.rswp` is only unique within a directory. For the central layout, build
the entry name from the file's absolute path with separators replaced by `%`,
e.g. `%Users%x%src%proj%main.go.rswp`, guarded by a `NAME_MAX` (255 byte) cap
that truncates the prefix and appends a hash of the full path so long paths stay
unique and deterministic.

Resolve symlinks before mangling so a file reached through a link shares a single
swap entry. `flush` already calls `scheme.Readlink`
(`internal/workspace/file.go:901`).

The mangling must be shared by `swapFileName` and `workspace.DefaultSwapFile`
(`internal/workspace/uri.go:48`), since the recovery prompts recompute the name
independently.

Upside: two workspace roots that reach the same file through different paths now
collide on the same swap entry, so the "already open by another session"
detection (`O_EXCL` → `ErrFileAlreadyOpen`, `internal/workspace/file.go:150`)
gets stronger.

### 3. Remote workspaces must keep the swap on the file's host

`initSwapFile` opens through `f.scheme.OpenFile`, and `schemeWorkspace.Load`
rejects a swap directory whose scheme, host or user differs from the workspace
(`internal/workspace/scheme_workspace.go:139`).

Keeping the URI path literally `~/.rune/swap/...` works for every scheme, because
`fileScheme.OpenFile` / `Rename` / `Remove` expand `~` on their own host
(`internal/workspace/file_scheme.go:190`). SSH and `rune://` workspaces therefore
get the remote `~/.rune/swap` for free.

Resolving the directory eagerly (mirroring `installDataDir`,
`internal/ide/workspace_handler.go:2361`) is the alternative, but it breaks for
files opened from foreign workspaces through `multi.loadExtraneous`
(`internal/workspace/multi_workspace.go:78`), whose host differs from the current
workspace's.

### 4. The swap directory must be created

`initSwapFile` uses `O_CREATE|O_EXCL` with no `MkdirAll`. `schemeapi.FileSystem`
exposes `MkdirAll`, so it must be called before the first `O_EXCL` create on a
given scheme.

### 5. Orphan swap files accumulate

Crashes leave entries in a now-global directory forever. Centralizing makes an
age-based reap or a `:swapfiles`-style listing command feasible. Out of scope
here, but worth a follow-up.

## Testing

New coverage:

- Config: `editor.directory` default, `"."`, a central path with `//`, a central
  path without `//`, and a malformed value recording `editor.directory` in
  `ideConfig.errors`.
- Both layouts run through the existing open / edit / flush / recover suites, so
  the sibling path is proven unchanged.
- Table-driven mangling tests: nesting, `%` in the original path, non-ASCII,
  symlink resolution, and `NAME_MAX` truncation determinism.
- Flush fallback: inject a scheme whose `Rename` returns `EXDEV`, then assert the
  file is saved correctly, the swap survives the save, and the original's mode
  and ownership are unchanged.
- Hard-linked and symlinked targets keep their links across a save on the
  fallback path.
- A crash simulated between "swap written" and "original written" recovers the
  full buffer.
- `~` expansion of the swap directory on a non-file scheme.
- Recovery round-trip under the central layout: crash, reopen, recover, force
  edit.
- Two workspaces reaching the same absolute path both observe
  `ErrFileAlreadyOpen`.

Existing tests that hardcode the sibling layout. They stay valid as coverage of
`editor.directory = "."`, but the ones that assert the *default* need a central
layout equivalent:

- `internal/workspace/uri_test.go` (`TestDefaultSwapFile`, `TestDefaultSwapDirectory`)
- `internal/workspace/file_test.go`
- `internal/workspace/workspacetest/scheme_suite.go`
- `internal/workspace/workspacetest/multi_workspace_test.go`
- `internal/workspace/workspacetest/manager_test.go`
- `internal/workspace/workspacessh/test/integration_test.go`
- `internal/handler/command/completer_test.go`
- `cmd/rune-agent/agent/agentools/agentools_test.go`, `cmd/rune-agent/extension/handler_test.go`

Validation: `make generate && make lint && make test`.

ernestrc:
Rune wrote the swap next to the file being edited, so every repository it touched
needed a *.rswp gitignore entry, the sibling entries polluted file listings and
search results, and opening a file in a directory the user cannot write failed
outright because the swap could not be created.
    
Saving no longer depends on renaming the swap over the original, which fails with
EXDEV once the two live on different filesystems. The rename stays as a fast path
and falls back to an in-place write, which also preserves the original's inode and
holds the swap's O_EXCL lock continuously across the save.
    
Closes #91
    
Signed-off-by: ernestrc <ernest@unstable.build>


internal/workspace/file.go
```diff
@@ -271,9 +265,16 @@ func (f *file) initFiles(filePath, swapDir string, readOnly bool) error {
...
-	swapDir, swapFileName := swapFileName(swapDir, filePath)
 	if f.swapFileName == "" {
-		f.swapFileName = swapFileName
+		var dir string
+		dir, f.swapFileName = swapFileName(swapDir, filePath)
+		swapDir = sharedSwapDir(dir, filePath)
+	}
+
+	if swapDir != "" {
+		if err := f.scheme.MkdirAll(swapDir, swapDirMode); err != nil {
```
zhengqunkoo: MkdirAll here solves p4 in https://github.com/unstablebuild/rune/issues/91 - initSwapFile is called a few lines later

ernestrc:
merged in bc62bbc209d18021b7fd4bb1800ade2a5e8a46bb

zhengqunkoo (commented):
Submitting half finished review

